PR #1, opened 2019-05-20
cleanup some build warnings

scholar-mage:


avivbeeri:
Hi there, thank you for contributing.

While I appreciate that some of the external libraries produce some spurious warnings, I'm not sure I like the idea of suppressing them by declaring them as system libraries.

I think a better solution would be to patch those libraries instead.

scholar-mage:
Hi. 
I suppose I have some mixed feelings about it, but it's good that we can have a discussion about that. 

If we plan on fixing them and getting those changes merged upstream, then I completely agree with you.

If we plan on maintaining separate patched versions for this project, well, I think that makes it harder to update those dependencies when the upstream author makes updates.

This patch was made with the thought that the upstream issues are upstream issues, and we'd probably rather focus our energy on building dome itself.

But aside from all that, I just wanted to reach out and say that I really like the idea and concept of the project. Wren is a cool language, and I've thought about building something like this myself.

Also, one issue I did run into is that dome currently complains if you try to run a wren file that's in a different directory. I tried to create a new game with dome as a submodule, and it refused to run my main file. I might poke around in there a bit, but haven't gotten to that yet.

I also started looking at memory leaks via valgrind. There's probably only about 5 or 6 of them during a clean exit. I managed to fix one of them so far. Still digging around on the others. I haven't looked at abnormal exits yet. I can post a gist later. 

- Brian

avivbeeri:
That's definitely an interesting way of looking at the problem. From that perspective, your solution makes much more sense.

I'm glad you reached out, and are finding value in the project so far. I'd be happy to discuss plans and ideas for DOME, perhaps across a different medium. I'm reasonably active in the wrenlang discord. I started this with the intention that others may want to have something similar. What platform(s) are you developing for? 

There's definitely still some work to do on the way paths are resolved, and what they are relative to. I tend to put the dome executable in a repo and build up the game's scripts from that location. I'm half-way to the goal of being able to bundle up a game's assets and code into a single executable file, so keeping everything self-contained is important.

I do try to keep on top of memory leaks, but I haven't run valgrind for a while. The exception handling definitely could use some work.

scholar-mage:
I primarily develop for Linux in C/C++/Python. I'm using various versions of Ubuntu. I have also done OSX and Windows development in the past, though I don't currently have an OSX machine that can run the current version of the OS. 

I might also have interest at someday in doing a port to the Ubuntu Touch phone or Librem 5 phone when it comes out. Nobody seems to have heard of either of those phone OSes, so I'd put that in the very-low priority bucket.

I'd also like to add gamepad support if it's not currently in there. I didn't try to test that, but only saw a module named keyboard.

Yeah I can hop on discord. I'm in the US/EDT time zone.

- Brian

avivbeeri:
Excellent! I primarily develop on OSX, so to do Linux or Windows builds requires me to use a virtual machine. I care mostly about PC builds, but I am interested in mobile ports, since SDL is available for a number of different platforms.

The input handling definitely needs expansion: Both mouse and gamepad options are on the roadmap. I've mostly used DOME for small projects so I haven't needed either of them yet.

I'm going to merge your pull request and we can continue this discussion elsewhere :)

Issue #3, opened 2019-05-30
Ogg files inside eggs cause an assertion failure


scholar-mage:
Eggs can contain ogg files, but it looks like the audio engine doesn't load them appropriately, and instead causes an assertion failure. This is reproducible on both Windows and Linux.

```$ ./dome 
Loading bundle /home/brian/projects/wren/gamepad/game.egg
Reading tar: ./main.wren
Reading from tar: ./main.wren
Loading module audio
Loading module input
Loading module graphics
Loading module point
Loading module random
Loading module ./test
Reading tar: ./test.wren
Reading from tar: ./test.wren
New Module
Loading module io
Reading tar: ./res/ship1.png
Reading from tar: ./res/ship1.png
Image loaded: res/ship1.png
Reading tar: ./res/ship2.png
Reading from tar: ./res/ship2.png
Image loaded: res/ship2.png
Reading tar: ./res/enemy.png
Reading from tar: ./res/enemy.png
Image loaded: res/enemy.png
Reading tar: ./res/heart-full.png
Reading from tar: ./res/heart-full.png
Image loaded: res/heart-full.png
Reading tar: ./res/heart-empty.png
Reading from tar: ./res/heart-empty.png
Image loaded: res/heart-empty.png
Reading tar: ./res/Laser_Shoot.wav
Reading from tar: ./res/Laser_Shoot.wav
Frequency: 44100 Hz
Samples: 4096
Channels: 1
Format: Signed 16 bit (LSB)
Audio loaded: res/Laser_Shoot.wav
Reading tar: ./res/Explosion.wav
Reading from tar: ./res/Explosion.wav
Frequency: 44100 Hz
Samples: 4096
Channels: 1
Format: Signed 16 bit (LSB)
Audio loaded: res/Explosion.wav
Reading tar: ./res/around-the-corner.ogg
Couldn't find ./res/around-the-corner.ogg in bundle, falling back.
dome: src/engine/audio.c:146: AUDIO_allocate: Assertion `data->length != UINT32_MAX' failed.
Aborted
```

scholar-mage:
Ok. This is actually caused by a missing `res/around-the-corner.ogg` in the stock game.egg file. Still, we probably shouldn't crash when an asset isn't found. It'd be better to provide a helpful error message instead.

avivbeeri:
Should be fixed now.

PR #15, opened 2019-05-31
Missing ogg file

scholar-mage:
This only partially address #3 by adding the missing ogg file to the game.egg. 

We still need to decide on an appropriate error handling strategy, so I will defer on that part.


scholar-mage:
Closed this because I noticed an incorrect owner/group in the egg file for this pull request. Will need to redo this one.

Issue #5, opened 2019-05-30
[Feature Request] A function call to exit the application


scholar-mage:
It'd be nice to have a function call that could be called from wren code to instruct DOME to exit / shutdown cleanly.

avivbeeri:
Done!

PR #16, opened 2019-10-22
Introduce a DOME module and an exit function

avivbeeri:
Implements request #5 

There is now a DOME module which can contain DOME-centric configuration
methods and properties. It also adds an `exit` method which can be used
by a game to exit the execution.

~~A caveat to this is that you can't quit a VM from within the C code,
so we need to wait for the execution to reach the core engine loop
before we can safely shutdown the Wren VM.~~

EDIT: We can use setjmp/longjmp to bypass the WrenVM execution and execute the DOME shutdown sequence, which cleans up memory safely.

### DIFF
--- a/main.wren
+++ b/main.wren
@@ -2,6 +2,7 @@ import "input" for Keyboard
 import "graphics" for Canvas, Color, ImageData, Point
 import "audio" for AudioEngine
 import "random" for Random
+import "dome" for Process
 import "./test"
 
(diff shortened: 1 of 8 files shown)


avivbeeri:
Because Wren doesn't allow for modules to be automatically imported, I'm considering changing the name of the module from `dome` to `process`, to be a more focused identifier.

PR #18, opened 2019-10-31
Audio Engine Improvements

### LINKED ISSUE #2: Attempting to play more than 4 sound channels at once results in the excess channel's playback being delayed
DOME doesn't advance the sample time position on channels which aren't played, so once a sound ends and the channel closes, the next sound plays from the beginning, rather than mid-way through.

avivbeeri:
This fixes #2 by making the audio channel list resizable.

We also moved to using the SDL callback which pushes mixing into a separate thread, so that audio won't be blocked by the game processing update loop.

### DIFF
--- a/src/modules/audio.c
+++ b/src/modules/audio.c
@@ -33,7 +33,6 @@ typedef struct {
 typedef struct AUDIO_ENGINE_t {
   SDL_AudioDeviceID deviceId;
   SDL_AudioSpec spec;
-  uint8_t* outputBuffer;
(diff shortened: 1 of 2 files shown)


src/modules/audio.c
```diff
@@ -1,11 +1,12 @@
-#define AUDIO_CHANNEL_MAX 8
+#define AUDIO_CHANNEL_START 2
```
avivbeeri: This should probably be bigger. Maybe 4?

PR #22, opened 2019-11-05
Performance

avivbeeri:
* Moving our timing point until after the SDL_PollEvent
* Switching from SDL_GetTicks to SDL_GetPerformanceCounter/SDL_GetPerformanceFrequency

This runs at a steady 58FPS with the example game.


src/main.c
```diff
@@ -286,16 +288,22 @@ int main(int argc, char* args[])
       goto cleanup;
     }
 
+    char buffer[20];
```
avivbeeri: I'd like this to be toggle-able, so we don't just draw it on every user's display.
Perhaps this is the start of a "debug display"?

Issue #23, opened 2019-11-05
Include a default manifest for itch.io


Sandvich:
Since I imagine a lot of DOME users will release on itch.io, it would be worthwhile providing a simple manifest for itch, as described [here](https://itch.io/docs/itch/integrating/manifest.html).

avivbeeri:
I'm not sure this is something the DOME project needs to provide, given how minimal the itch manifests can be.

Issue #4, opened 2019-05-30
IMAGE_allocate will attempt to load a nonexistent file, causing a crash


scholar-mage:
Dome will attempt to load invalid resource paths, which causes a segfault. For example, rename ship1.png to something else, and you'll get the following crash:

```
Loading module audio
Loading module input
Loading module graphics
Loading module point
Loading module random
Loading module io

Thread 1 "dome" received signal SIGSEGV, Segmentation fault.
0x000055555556ed75 in stbi__get8 (s=0x7fffffffbb30) at src/include/stb_image.h:1409
1409	      return *s->img_buffer++;
(gdb) where
#0  0x000055555556ed75 in stbi__get8 (s=0x7fffffffbb30) at src/include/stb_image.h:1409
#1  0x0000555555575e26 in stbi__get_marker (j=0x5555576b5500) at src/include/stb_image.h:2660
#2  0x000055555557866c in stbi__decode_jpeg_header (z=0x5555576b5500, scan=1)
    at src/include/stb_image.h:3098
#3  0x000055555557bc04 in stbi__jpeg_test (s=0x7fffffffbb30) at src/include/stb_image.h:3750
#4  0x000055555556dc4f in stbi__load_main (s=0x7fffffffbb30, x=0x5555576b54d8, y=0x5555576b54dc, 
    comp=0x5555576b54e0, req_comp=4, ri=0x7fffffffbadc, bpc=8) at src/include/stb_image.h:980
#5  0x000055555556e0c9 in stbi__load_and_postprocess_8bit (s=0x7fffffffbb30, x=0x5555576b54d8, 
    y=0x5555576b54dc, comp=0x5555576b54e0, req_comp=4) at src/include/stb_image.h:1088
#6  0x000055555556e7ec in stbi_load_from_memory (buffer=0x0, len=1431883287, x=0x5555576b54d8, 
    y=0x5555576b54dc, comp=0x5555576b54e0, req_comp=4) at src/include/stb_image.h:1226
#7  0x0000555555567f5e in IMAGE_allocate (vm=0x5555575c2c20) at src/engine/image.c:19
#8  0x000055555558b3ff in runInterpreter ()
#9  0x0000555555569deb in main (argc=1, args=0x7fffffffde58) at src/main.c:172
```

avivbeeri:
This is fixed for asset loading, but will need to be revisited when async loading is introduced.

PR #27, opened 2019-12-23
Bug fixes

avivbeeri:
Allow users to control elements of how the DOME game loop works, by having a setting to enable vsync and "lockstep" game loops.

### DIFF
--- a/docs/modules/dome.md
+++ b/docs/modules/dome.md
@@ -24,8 +24,12 @@ Allows you to programmatically close DOME. This command behaves a little differe
 
 ### Static Fields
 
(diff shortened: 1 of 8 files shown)


avivbeeri:
This branch also completes fixes for #4 and #23 
Assets were recently changed to be loaded using FileSystem.loadSync, but now it also aborts the fiber if a file can't be found on the filesystem, rather than returning null and then being parsed in an invalid manner. 

The async IO api (which is still highly unstable) doesn't have a method for reporting errors yet, so if you were to somehow load files asynchronously using it, you would still segfault.

PR #39, opened 2020-01-20
Rotation and other transforms

avivbeeri:
Still needs docs, as well as filling out all the other transformation parameters.


avivbeeri:
Closed, as [#40] is deemed less error prone. Ideas from this could be considered for future versions.

PR #42, opened 2020-01-30
Lazy module load

avivbeeri:
We don't init gamecontroller and audio systems unless required.

Includes other tweaks aimed at reducing the overhead of loading library data at startup.


src/vm.c
```diff
@@ -71,19 +71,28 @@ internal WrenForeignMethodFn VM_bind_foreign_method(
     bool isStatic,
     const char* signature) {
 
+  // For convenience, concatenate all of the method qualifiers into a single
```
avivbeeri: I need to add a disclaimer/copyright notice here, since this block was ripped from the Wren test codebase.

src/util/generateEmbedModules.sh
```diff
@@ -19,14 +19,14 @@ declare -a opts=(
 rm ../modules/modules.inc 2> /dev/null
 touch ../modules/modules.inc
 
-rm ../modules/modulemap.inc 2> /dev/null
+rm ../modules/modulemap.c.inc 2> /dev/null
```
avivbeeri: This typo malformed the map includes file somewhat since it wasn't being deleted between `make`s.

PR #44, opened 2020-02-09
Gif record

avivbeeri:
The current gif recording system (`--record=gifName.gif`) is rather intensive on the main thread, so this PR offloads the file IO and pixel-channel-rearrangement to its own thread.

Unfortunately, it still struggles to keep a good frame-rate, depending on the size of the canvas and number of colours in use, so I don't recommend it for serious work.


avivbeeri:
Needs refactoring to put the thread handling inside the engine lifecycle.

PR #59, opened 2020-03-24
added version module

clsource:
useful for displaying and checking the current dome version
(maybe for api changes in the future, games can check if the dome version is compatible)

example usage

```js
import "dome" for Version, Process, Window
import "input" for Keyboard

class Game {

    static init() {
     Version.print()
     Version.isAtLeast(1)
    }
    
    static update() {
      if(Keyboard.isKeyDown("escape")) {
        Process.exit()
      }
    }

    static draw(dt) {
    }
}

```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -7,13 +7,15 @@
 
 #ifndef DOME_VERSION_NUMBER
 // The same as DOME_VERSION but only in numeric form
-// major.minor.patch.status
+// major.minor.patch-status
 // status must be one of these
-// 0 = alpha, 1 = beta, 2 = rc, 3 = stable, 4 = hotfix, 5 = reserved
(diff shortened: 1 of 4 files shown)


avivbeeri:
I like the idea of a Version API, but I'm not sure I like this implementation. At the moment, I actually set the DOME_VERSION manually as a string during the final build compilation, so I'd have to change my build processes.

Being able to do fine-grained comparisons is useful, so I'll consider some alternatives.

clsource:
Does not need to change the build process to update the build number.

As shown in main.c

https://github.com/avivbeeri/dome/blob/56bebab0fbf5e5a8b21f74f6901d77a277688fb4/src/main.c#L19

You can manually set the version number using

`DOME_VERSION_NUMBER` constant 
before release.

It's basically the same process :).

clsource:
I belive a better implementation maybe how wren does it

here
https://github.com/wren-lang/wren/blob/master/src/include/wren.h

PR #61, opened 2020-03-25
fixed negative radius bug in circles

### LINKED ISSUE #60: Some API's don't handle negative values correctly
For example, trying to draw a circle with a negative radius results in a segfault or worse.

clsource:
fixes #60


avivbeeri:
I don't consider this a complete fix, because #60 is about checking all the graphics routines for negative-range bugs. Most likely, something similar applies to rectfill (negative width/height) and possibly ellipse/lines (although these are less likely).

clsource:
Ok. I believed that was only concerning circles. May then be needed to check the other graphics

clsource:
futher investigation shows that pico8 and tic80 does not draw a circle with negative radius

```lua
cls(0)
circ(20, 50, 10, 9)
circ(20, 70, -1, 2)
```

imagen

PR #62, opened 2020-03-30
Improve importing modules

frarees:
- Priorize builtin modules over files
- Fix loading from game.egg
- Fix loading resources from macOS app bundle
- `./` on custom modules is optional

### DIFF
--- a/src/engine.c
+++ b/src/engine.c
@@ -113,51 +113,39 @@ ENGINE_writeFile(ENGINE* engine, char* path, char* buffer, size_t length) {
 
 internal char*
 ENGINE_readFile(ENGINE* engine, char* path, size_t* lengthPtr) {
-  if (engine->tar != NULL) {
-    char pathBuf[PATH_MAX];
-    strcpy(pathBuf, "\0");
-    if (strncmp(path, "./", 2) != 0) {
-      strcpy(pathBuf, "./");
-    }
-    strcat(pathBuf, path);
-    mtar_header_t h;
-    int success = mtar_find(engine->tar, pathBuf, &h);
-    if (success == MTAR_ESUCCESS) {
-      ENGINE_printLog(engine, "Reading from bundle: %s\n", path);
-      char* file = readFileFromTar(engine->tar, pathBuf, lengthPtr);
-      if (file == NULL) {
-        ENGINE_printLog(engine, "Error: Couldn't read the data from the bundle.");
-      }
-      return file;
-    } else if (success != MTAR_ENOTFOUND) {
-      ENGINE_printLog(engine, "Error: There was a problem reading %s from the bundle.\n", pathBuf);
-      return NULL;
-    }
-    ENGINE_printLog(engine, "Couldn't find %s in bundle, falling back.\n", pathBuf);
+  char pathBuf[PATH_MAX];
+
+  if (strncmp(path, "./", 2) == 0) {
+    strcpy(pathBuf, path + 2);
+  } else {
+    strcpy(pathBuf, path);
   }
 
-  char* fullPath = NULL;
+  if (engine->tar != NULL) {
+    ENGINE_printLog(engine, "Reading from bundle: %s\n", pathBuf);
+
+    char* file = NULL;
+    int err = readFileFromTar(engine->tar, pathBuf, lengthPtr, &file);
+    if (err == MTAR_ESUCCESS) {
+      return file;
+    }
+
+    if (DEBUG_MODE) {
+      ENGINE_printLog(engine, "Couldn't read %s from bundle: %s. Falling back\n", pathBuf, mtar_strerror(err));
+    }
+  }
+
   if (path[0] != '/') {
-    char* base = BASEPATH_get();
-    fullPath = malloc(strlen(base)+strlen(path)+1);
-    strcpy(fullPath, base); /* copy name into the new var */
-    strcat(fullPath, path); /* add the extension */
-  } else {
-    fullPath = path;
+    strcpy(pathBuf, BASEPATH_get());
+    strcat(pathBuf, path);
   }
-  if (!doesFileExist(fullPath)) {
-    if (path[0] != '/') {
-      free(fullPath);
-    }
+
+  if (!doesFileExist(pathBuf)) {
     return NULL;
-  } else {
-    ENGINE_printLog(engine, "Reading from filesystem: %s\n", path);
-    char* data = readEntireFile(fullPath, lengthPtr);
-    if (path[0] != '/') {
-      free(fullPath);
-    }
-    return data;
   }
+
+  ENGINE_printLog(engine, "Reading from filesystem: %s\n", pathBuf);
+  return readEntireFile(pathBuf, lengthPtr);
 }
 
 internal int
--- a/src/main.c
+++ b/src/main.c
@@ -257,63 +257,57 @@ int main(int argc, char* args[])
   {
     char* defaultEggName = "game.egg";
     char* mainFileName = "main.wren";
-    char* fileName = defaultEggName;
 
     char* base = BASEPATH_get();
     char* arg = optparse_arg(&options);
+
+    char pathBuf[PATH_MAX];
+
+    char* fileName = NULL;
+
     if (arg != NULL) {
-      fileName = arg;
-    } else {
-      fileName = defaultEggName;
-    }
-
-    size_t baseLen = strlen(base);
-    size_t fileNameLen = strlen(fileName);
-    char* pathBuf = calloc(baseLen + fileNameLen + 1, sizeof(char));
-    strcpy(pathBuf, base);
-    strcat(pathBuf, fileName);
-    // Re-target basepath to the argument given
-    if (isDirectory(pathBuf)) {
-      BASEPATH_set(pathBuf);
-      fileName = defaultEggName;
-      fileNameLen = strlen(fileName);
-      arg = NULL;
-    } else {
-      BASEPATH_set(dirname(pathBuf));
-    }
-    base = BASEPATH_get();
-    baseLen = strlen(base);
-
-    if (isDirectory(pathBuf)) {
-      pathBuf = realloc(pathBuf, (baseLen + fileNameLen + 1) *  sizeof(char));
       strcpy(pathBuf, base);
-      strcat(pathBuf, defaultEggName);
+      strcat(pathBuf, arg);
+      if (isDirectory(pathBuf)) {
+        BASEPATH_set(pathBuf);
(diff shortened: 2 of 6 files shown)


avivbeeri:
Note: This change breaks existing .egg files made with NEST (which uses the `./` prefix on files).

src/vm.c
```diff
@@ -88,15 +88,18 @@ internal WrenForeignMethodFn VM_bind_foreign_method(
 internal char* VM_load_module(WrenVM* vm, const char* name) {
   ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
   MAP moduleMap = engine->moduleMap;
-  if (strncmp("./", name, 2) != 0) {
-    if (DEBUG_MODE) {
-      ENGINE_printLog(engine, "Loading module %s\n", name);
```
avivbeeri: Why was this logging removed?

frarees: In the og file, there are two possible returns: first, look up if the module is not a custom one by matching `./` and debug log the (builtin) module, and second, look up the filesystem otherwise and log the module. 

What I did in my commit is get rid of one of those logs and make logging happen before any of this, so that we have one log, no matter where we get the module from.

frarees:
> Note: This change breaks existing .egg files made with NEST (which uses the `./` prefix on files).

My initial test, providing a folder: 

```
frarees@MBP ivokdev % nest -z -o game.egg -- gamedata
Bundling: ./gamedata/tilemap.png
Bundling: ./gamedata/tilemap.wren
Bundling: ./gamedata/main.wren
Created bundle game.egg.
frarees@MBP ivokdev % tar tvf game.egg 
-rw-rw-r--  0 0      0        1459 Jan  1  1970 /tilemap.png
-rw-rw-r--  0 0      0         269 Jan  1  1970 /tilemap.wren
-rw-rw-r--  0 0      0         326 Jan  1  1970 /main.wren
```

A second test, given your results, providing just file paths:

```
frarees@MBP ivokdev % nest -z -o game.egg -- gamedata/*.wren
Bundling: ./gamedata/main.wren
Bundling: ./gamedata/tilemap.wren
Created bundle game.egg.
frarees@MBP ivokdev % tar tvf game.egg 
-rw-rw-r--  0 0      0         322 Jan  1  1970 ./gamedata/main.wren
-rw-rw-r--  0 0      0         281 Jan  1  1970 ./gamedata/tilemap.wren
```

So there's clearly differences. Since microtar seems to need the exact string (I'm not very familiar with the library, but seems like it can't handle path conversions or expansions), I suggest to:

1. Make sure nest produces consistent paths, be it root-based `/`, rel-path based `./` or non-prefixed. Non-prefixed makes the most sense to me. If we need further prefixing as an implementation detail, we could prepend it in code.
2. Decide if/make sure we are able to load game.egg files generated before the changes. Since it seems game.egg files can be created with `./` and `/` prefixes (as demonstrated before), if we want to maintain compatibility we would need to call microtar with both `/` and `./` variants. Personally, I think it makes sense to not allow backwards compatibility, since nest is still in development and I guess we can afford users re-generating egg files (maybe add functionality on nest to sanitize egg files?)

frarees:
There are still cases not covered by the PR, such as loading from a specified directory. I'll be looking at that too

avivbeeri:
So the current state (as of this comment) is nice (I appreciate the relative simplicity), but it will break existing game code which uses imports prefixed with `./`, since those will no longer be found in egg files, and the check for prefixes has been removed.

frarees:
>  it will break existing game code

Indeed. I plan on working on backwards compatibility, now that I have something cleaned up and working. That is, if it makes sense and we agree to go forward with all these dome+nest path related changes.

frarees:
Given how game.egg can have both `./` and `/` prefixes, we can:

1. Try both variants when asking for a file on microtar. Con: overhead for "legacy eggs".
2. Include a tiny manifest in .egg where we store (at least) the egg version, so that we know what we're dealing with.

I guess 2 makes most sense for a robust and ongoing tool. Should we go with that?

avivbeeri:
Alternatively, don't allow '/' prefixes at all. DOME uses '/' prefixing to assume absolute paths, so maybe it shouldn't be using the .egg for those at all.

I do like the idea of a small manifest for future considerations, but as a safety/security thing, DOME shouldn't be able to accidentally load it as a wren module.

frarees:
> Alternatively, don't allow '/' prefixes at all

nest currently prepends `/` when you build a folder e.g. `nest -z -o game.egg -- folder` (see nest's main.c:44), so if we're ok with not supporting legacy eggs being built with folders (for the lack of a better term), point 1. would be:

1. Fetch `file` and (fallback to) `./file` via microtar. Con: overhead for "legacy eggs".

avivbeeri:
Oh problematic (I haven't looked much at nest's implementation since I wrote it like a year ago).

If we have a framework for supporting the legacy egg formats, then we can handle both, although it does complicate things a bit in an already complicated area of code.

I think I need a better view of who is using the EGG file format currently. If it's mostly just me, then we can probably assume it to be totally "unstable" and start again.

frarees:
I've cleaned up the PR so that everything is contained within one commit. The core functionality is there. I've tested the changes with every entry on the jam (and some more small tests toying with mainstream nest and https://github.com/avivbeeri/nest/pull/2).

Changes have been tested on macOS Catalina and RG350 (Opendingux, linux-based), haven't tried Window out. 

We need to update the docs to reflect the changes, check code conventions for consistency, and I guess we're good to go. 

In case we don't need backwards compatibility with previous egg files (and assuming we're OK with https://github.com/avivbeeri/nest/pull/2) I could get rid of the extra compares.

frarees:
Found a small issue. Importing a builtin module (e.g. `import "random" for Random`) does log a misleading error:

```
Loading module random
Reading from bundle: random.wren
Error: There was a problem reading random.wren from the bundle.
```

Although it works.

avivbeeri:
Yeah, I spotted that also. The built in modules do make this slightly
awkward.

frarees:
While I was working on this PR, I've noticed the startup time has gotten bigger. I'm not sure if it's related specifically to the changes on this PR (if so... how? any clue?). I tested your rhythm jam entry (https://avivbeeri.itch.io/rhythmdrive) and could see a noticeable difference between my built (release) dome and the one bundled.

src/engine.c
```diff
@@ -113,51 +113,40 @@ ENGINE_writeFile(ENGINE* engine, char* path, char* buffer, size_t length) {
...
-        ENGINE_printLog(engine, "Error: Couldn't read the data from the bundle.");
-      }
+    ENGINE_printLog(engine, "Reading from bundle: %s\n", pathBuf);
+
+    char* file = NULL;
+    int err = readFileFromTar(engine->tar, pathBuf, lengthPtr, &file);
+    if (err == MTAR_ESUCCESS) {
       return file;
-    } else if (success != MTAR_ENOTFOUND) {
+    } else if (err == MTAR_ENOTFOUND) {
```
avivbeeri, line 131: Turns out, microtar doesn't always use this error code. Running it on the spaceshooter example results in `MTAR_ENULLRECORD` when trying to read config.txt (a file which is explicitly not included in the `game.egg` to test for fallback IO.) I think we are safe to check for both.

frarees: We might want to get rid of bundle to fs fallback altogether (see 8292dc9)

avivbeeri: The fallback is intentional, because without it, you can't have local save files or configuration/preference files.

frarees: Ahh I completely missed that one. What if we have two possible APIs? One to read from assets, so potentially always a RO operation (e.g. bundle, fs, romfs... you name it. We could later use a layer here with something like physfs), and other to read/write from data on the fs?

avivbeeri: I would quite like to avoid that, for a couple reasons.
1) It complicates the FS api, which is against the design philosophy (and may be considered a breaking change
2) Impractical usage: The programmer likely won't be using a .egg until the end of development, and so once they switch to `game.egg`, they'd have to change their code.

The original design makes the bundle transparent, for easier distribution.

frarees: I see. Then I guess we could just fallback to fs if tar reading fails for whatever reason, instead of just on `MTAR_ENOTFOUND`?

avivbeeri: I think that's a suitable solution.

src/engine.c
```diff
@@ -130,7 +130,6 @@ ENGINE_readFile(ENGINE* engine, char* path, size_t* lengthPtr) {
       return file;
     } else {
       ENGINE_printLog(engine, "Error reading %s: %s\n", pathBuf, mtar_strerror(err));
```
avivbeeri, line 132: We probably want to use the original fall-back message, rather than the bundle-specific one.

avivbeeri:
Tested this on Windows, seems sufficient

src/main.c
```diff
@@ -257,63 +257,52 @@ int main(int argc, char* args[])
...
-      BASEPATH_set(dirname(pathBuf));
-    }
-    base = BASEPATH_get();
-    baseLen = strlen(base);
+    char pathBuf[PATH_MAX];
+
+    char* fileName = NULL;
+
+    if (arg != NULL) {
+      strcpy(pathBuf, arg);
```
avivbeeri, line 269: This needs to be strcat, and we need a `strcpy(pathBuf, base)` here, because otherwise on some OS, a relative file argument like `../` will fail.

frarees: Updated. Can you verify it works on debian now?

avivbeeri: So this works to solve the issue if you run `./dome ../`, but it doesn't work for `./dome ../game.egg`. I'm trying to pin down the root cause.

src/main.c
```diff
@@ -257,63 +257,53 @@ int main(int argc, char* args[])
...
+    char* fileName = NULL;
+
+    if (arg != NULL) {
       strcpy(pathBuf, base);
-      strcat(pathBuf, defaultEggName);
+      strcat(pathBuf, arg);
+      if (isDirectory(pathBuf)) {
+        BASEPATH_set(pathBuf);
+      } else {
+        fileName = basename(pathBuf);
```
avivbeeri, line 274: Found the problem. `basename`/`dirname` can modify the string passed in, so we need to use a throwaway copy of `pathBuf` for each call here.

frarees: Does 0c95335 solve the issue?

avivbeeri: Seems like it! Nice work!

avivbeeri:
This seems to work for Windows, Mac, Linux, in all the crazy cases I can think of, so we are probably good to go! Squash it and I can merge :)

PR #78, opened 2020-08-26
implemented ImageData.pget, which is useful for 3D texture rendering

avivbeeri:
Very simply, provides `ImageData.pget` which can be used to retrieve the color value of a pixel in ImageData. This would be useful for some side projects, where the image loaded in isn't directly blitted to the display, but is instead used for other purposes.

This also changes the `toString` of the Color class to be more human-readable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 DOME-out.log
 .mode
+.arch
 dome
 nest
 libwren.a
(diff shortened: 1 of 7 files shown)


avivbeeri:
I _think_ that this PR's feature set is complete now, since you can create, modify and read "ImageData", if you want to.

PR #82, opened 2020-09-24
[WIP] added json support + tests

clsource:
This is a small test for adding json support to dome.

follows a Json Stream approach using https://github.com/skeeto/pdjson

```js
import "dome" for Json

json = Json.fromFile("complex.test.json").json

if(!(json is Map)) {
    System.print("JSON must be a Map")
    System.print(json)
    Process.exit(-1)
}

if(json["uid"] != "76-buteo-albigula") {
    System.print("uid not found")
    System.print(json)
    Process.exit(-1)
}

if(json["name"]["latin"] != "Buteo albigula") {
    System.print("name.latin not found")
    System.print(json)
    Process.exit(-1)
}
```


clsource:
ok I will rework this to be more simpler to merge

PR #83, opened 2020-09-26
Add Json support

clsource:
More polished version of #82 

- [x] add json support
- [x] add docs


docs/index.md
```diff
@@ -5,7 +5,7 @@ DOME is a framework for making 2D games using the [Wren programming language](ht
 
 Get help in the [DOME Discord Server](https://discord.gg/Py96zeH)!
 
-Join us for [DOMEjam](https://itch.io/jam/domejam) between March 13th and March 23rd!
+Join us for [DOMEjam](https://itch.io/jam/domejam) between March 13th and March 23rd!.
```
avivbeeri, line 8: I probably need to remove this line entirely, since it's long past!

docs/modules/json.md
```diff
@@ -0,0 +1,110 @@
...
+
+* [Json](#json)
+* [JsonOptions](#jsonoptions)
+* [JsonError](#jsonerror)
+
+## Json
+
+### Constructor
+
+#### `constructor parse(value:String, options:Num)`
```
avivbeeri, line 18: Is the average user expected to directly call this constructor, or use the static `parse` and `fromFile` methods?

docs/modules/json.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Static Methods
+
+#### `static parse(value:String): Json`
+Returns a new _Json_ object with default options.
+
+#### `static fromFile(path:String): Json`
+Reads the contents of a file in `path` and returns a new _Json_ object with default options.
+
+#### `static fromFile(path:String, options:Num): Json`
```
avivbeeri, line 57: I prefer to combine overloaded methods, and describe the defaults, rather than separate entries for each.
```
#### `static fromFile(path: String): Json`
#### `static fromFile(path: String, options: Num): Json`
Reads the contents of the file located at `path` and returns a new Json object with the specified options. If no options are provided, `JsonOptions.NIL` is assumed.
```

src/modules/json.c
```diff
@@ -0,0 +1,142 @@
+// JSON Parser Events
+const char json_typename[][64] = {
+    [JSON_ERROR]      = "com.domeengine.json:JSON_ERROR",
```
avivbeeri, line 3: I like this style of specifying errors. Wren and DOME's error handling isn't particularly powerful, but this sort of provides a structure.

src/modules/json.c
```diff
@@ -0,0 +1,142 @@
...
+
+internal void
+JSON_pos(WrenVM * vm) {
+  wrenSetSlotDouble(vm, 0, json_get_position(jsonStream));
+}
+
+internal void
+JSON_next(WrenVM * vm) {
+  enum json_type type = json_next(jsonStream);
+  switch (type) {
```
avivbeeri, line 61: The only benefit to using a switch statement here is to deal with a potential `default` error case, otherwise, you can probably do away with it.

src/modules/json.c
```diff
@@ -0,0 +1,142 @@
...
+  "\\" // A backslash.
+  "\a" // Alarm beep. (Who uses this?)
+  "\b" // Backspace.
+  "\f" // Formfeed.
+  "\n" // Newline.
+  "\r" // Carriage return.
+  "\t" // Tab.
+  "\v" // Vertical tab.
+  */
+  if(strcmp(value, "\0") == 0) {
```
avivbeeri, line 108: Wren strings are unicode, so we should use `utf8cmp` for these checks, just in case.

src/modules/json.wren
```diff
@@ -0,0 +1,252 @@
...
+
+    end() {
+        stream_end()
+    }
+
+    process(event) {
+
+        _lastEvent = event
+
+        if(event == isBoolTrue || event == isBoolFalse) {
```
avivbeeri, line 81: As a code-style thing, I prefer spaces after `if` and `while` keywords in the DOME codebase.

avivbeeri:
Just checking so that I understand how this works: the pdjson library returns a list of parsing events, which we are then processing on the Wren side to build up a Wren-based representation?

I assume the reason not to do this entirely on the C side is that we can't easily build up maps at the moment? 

Is there a good code-example of this module in use I can look at?

clsource:
> Just checking so that I understand how this works: the pdjson library returns a list of parsing events, which we are then processing on the Wren side to build up a Wren-based representation?
> I assume the reason not to do this entirely on the C side is that we can't easily build up maps at the moment?

Yes. My first approach was to parse everything in C. But I couldn't transform the final structure to a Wren Map.
The upside of this event-based approach is that you could read heavy jsons and build them on the fly.


> Is there a good code-example of this module in use I can look at?
The way APIs are designed and used is one of the most critical parts of how I design what goes into DOME, so we might have to change that here to better align with the project goals.

The old PR contained some example tests. https://github.com/NinjasCL-archive/dome/blob/ddb603c2dcdfa341058d1a32674a1d608d25208e/test/json/json.test.wren

But since there is no standardized testing process at the moment I just omitted the tests.
I am ok to redesign the APIs to be more DOME like.

docs/modules/json.md
clsource: Normally the expected way is using the static methods since they are a factory pattern using default options.
The construct is the base for all the other static methods.

avivbeeri:
We might want to wait for Wren 0.4.0 so that this can be handled from the C side.
I'll need to do a bit of thinking about how I'd want to handle JSON in DOME.

```
var obj = JSON.parse("string") // or JSON.fromFile(path)
System.print(obj["version"]["minor"])
var output = JSON.stringify(obj)
```
This would be my first impulse for API design, coming from a javascript-heavy background. Errors would always abort. We don't need optional escaping at the moment, since there's limited uses for the input.

I don't particularly like the map-chaining approach aesthetically (a lot of extra brackets and quotes), but it's the easiest way to manipulate the data and change it's shape for output.

clsource:
> I don't particularly like the map-chaining approach aesthetically (a lot of extra brackets and quotes), but it's the easiest way to manipulate the data and change it's shape for output.

Yes. But always the user can create a wrapper class for easier output.

Example Wrapper for the https://dog.ceo/api/breeds/list in Swift

```swift
import Foundation

public struct DogBreed: Codable {
    let names: [String]?
    let status: String?
    
    enum CodingKeys: String, CodingKey {
        case names = "message"
        case status
    }
}
```

> I'm not a particular fan of stringify. encode/decode is a common pairing for this sort of API (looking at lua and erlang at least). Might think about that a bit more.

Well im no fan of stringify either because is more verbose. encode/decode is used in _Swift_ too. 

```swift 
try! JSONDecoder().decode(DogBreed.self, from: jsonData)
```

But _Python_ likes `dumps()`. https://docs.python.org/3/library/json.html

avivbeeri:
`dumps` sounds really weird to me. Shorter, but it sounds immature to me :P encode/decode is probably the right answer, although remembering which one does which function isn't easy, since you could argue both functions are forms of encoding.

If I wanted a read-only solution (although I don't), what I'd have is something like:

```
var obj = JSON.fromFile("file.json")
System.print(obj["root.objects[4].name"])
```

But manipulating that kind of API to mutate, add, remove, iterate, etc isn't easy. Plus having to write a parser for that kind of pathing.

clsource:
What about `toString()`?

```js
import "json" for JSON
var obj = JSON.parse("string") // or JSON.fromFile(path)
System.print(obj.json["version"]["minor"])
var output = obj.toString() // or JSON.toString(obj.json)
```

avivbeeri:
I worry that `toString()` would be confusing, since every object has a
`toString` property, although it's at least clear about what the method
does.

I'm trying to avoid having a specific JSON instance object, so `obj.json[]`
isn't ideal. I'd like to try and stick to the one static JSON class, which
resolves to pure Wren objects.

clsource:
Ok how about this API?

```swift
Json.fromString(value:String) -> Map
Json.parse(value:String) -> Map // Convenience method for Json.fromString
Json.fromFile(path:String) -> Map

Json.encode(value:Map) -> String
Json.save(path:String, value:Map) -> Void // Will encode and save the result to a file path
```

avivbeeri:
I like API simplicity, but the naming is bugging me because there's basically no symmetry, which will make remembering it problematic. Let's go with this:
```
# File IO, mirroring the FileSystem class
JSON.load(path: String) -> <Map, List, Num, String, Null>
JSON.save(path: String, value: <Map, List, Num, String, Null>) -> Void

# General Manipulation - borrowed from other languages/apis
JSON.encode(value: <Map, List, Num, String, Null>) -> String
JSON.decode(value: String) -> <Map, List, Num, String, Null>
```

clsource:
Sounds good.
Is there any preference for using capitalized _JSON_ instead of _Json_?.

does It contradicts somehow the _Wren_ naming convention of Dome?

avivbeeri:
I've never known how to handle this with acronyms, but I always captialise
JSON when I type it.

We could allow both because of variable name exporting, but I'd like to
settle one way or another for documentation.

avivbeeri:
Let's canonically call it `Json`, to keep consistency with other APIs, but add `var JSON = Json` so I can have my cake and eat it also.

avivbeeri:
This PR hasn't moved for a while, can you get it into working order as discussed above?

clsource:
ok

clsource:
Test file `test.json`

```json
{
    "success":true,
    "file":true
}
```

`main.wren`
```js
import "json" for Json

class Game {
    static init() {
      System.print(Json.encode("Hello"))
      System.print(Json.decode("{\"success\":true}"))
      System.print(Json.load("test.json"))
      System.print(Json.save("test2.json", {"success":true}))
    }
    static update() {}
    static draw(dt) {
    }
}
```

src/modules/json.wren
```diff
@@ -0,0 +1,245 @@
...
+                _lastEvent = event
+                if (event == isObjectEnd || event == isDone || event == isError) {
+                    break
+                }
+                elements[this.value] = process(next)
+            }
+            return elements
+        }
+
+        if (event == isError) {
```
avivbeeri, line 123: It turns out that this only fires if the event occurs in the top level of an object. If an error is found inside a nested object/list/etc this won't be reached. We should be able to abort and jump out to the error handling case as soon as the error is discovered.

src/modules/json.wren
```diff
@@ -0,0 +1,245 @@
+import "io" for FileSystem
+
+class JsonOptions {
+    static NIL { 0 }
+    static ESCAPE_SLASHES { 1 }
+    static ABORT_ON_ERROR { 2 }
```
avivbeeri, line 6: I believe that ABORT_ON_ERROR should be a default behaviour, and disabled if necessary.

src/modules/json.wren
```diff
@@ -0,0 +1,245 @@
...
+
+    static shouldAbort(options) {
+        return (options == JsonOptions.ABORT_ON_ERROR || options == (JsonOptions.ABORT_ON_ERROR | JsonOptions.ESCAPE_SLASHES))
```
avivbeeri, line 9: This can be simplified to: `return options & JsonOptions.ABORT_ON_ERROR != 0`, which will be much more scalable, if extra options are added in the future.

src/modules/json.wren
```diff
@@ -0,0 +1,245 @@
...
+                }
+                elements[this.value] = process(next)
+            }
+            return elements
+        }
+
+        if (event == isError) {
+            _error = JsonError.new(lineno, pos, error_message, true)
+            if (JsonOptions.shouldAbort(_options)) {
+                Fiber.abort("JSON Error: line %(lineno) pos %(pos): %(error_message)")
```
avivbeeri, line 126: Please change this from `JSON Error: ` to `JSON error - ` so that the Runtime Error output looks a bit less repetitive.

src/modules/json.c
```diff
@@ -0,0 +1,132 @@
...
+  } else if (utf8cmp(value, "\v") == 0) {
+    result = "\\v";
+  } else if (utf8cmp(value, "/") == 0) {
+    // Escape / (solidus, slash)
+    // https://stackoverflow.com/a/9735430
+    // The feature of the slash escape allows JSON to be embedded in HTML (as SGML) and XML.
+    // https://www.w3.org/TR/html4/appendix/notes.html#h-B.3.2
+    // This is optional escaping. Disabled by default.
+    // use JsonOptions.ESCAPE_SLASHES option to enable it
+    if(options == JSON_DOME_ESCAPE_SLASHES ||
```
avivbeeri, line 125:
```
if ((options & JSON_DOME_ESCAPE_SLASHES) != 0) {
```

src/vm.c
```diff
@@ -291,6 +291,16 @@ internal WrenVM* VM_create(ENGINE* engine) {
...
 
+  // Json
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.stream_begin(_)", JSON_streamBegin);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.stream_end()", JSON_streamEnd);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.next", JSON_next);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.value", JSON_value);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.error_message", JSON_error_message);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.lineno", JSON_lineno);
+  MAP_addFunction(&engine->moduleMap, "json", "JsonStream.pos", JSON_pos);
+  MAP_addFunction(&engine->moduleMap, "json", "static JsonStream.escapechar(_,_)", JSON_escapechar);
```
avivbeeri, line 302: The C methods here "belong" to the JsonStream class, so they should be prefixed as `JSON_STREAM_`, rather than `JSON_`.

avivbeeri (changes requested):
A few changes based on testing. 

Please also fix the indentation in the `src/modules/json.wren` file to be 2-spaced.

src/modules/json.wren
clsource: In my tests pdjson would try to interpret the json the best it can. So even if the json is invalid it would parse it anyway.

https://github.com/skeeto/pdjson

"streaming" mode. When a JSON "stream" contains multiple JSON objects (optionally separated by JSON whitespace), the default behavior of the parser is to allow the stream to be "reset," and to continue parsing the stream.

clsource: Example 
```json
{
    "this can make an error":{
        "nested": {
            "error": true // This is invalid json
        }
        // But it will be considered anyway since the streaming support
        "nested2":{
            "error": falsc
        }
    }
}
```

It would be parsed as:

```{this can make an error: {nested: {error: true}}}```

File: src/modules/json.wren (hunk 9 above)
Comment (clsource) [reply src/modules/json.wren:6]: Ok

File: src/modules/json.wren (hunk 10 above)
Comment (clsource) [reply src/modules/json.wren:9]: done

File: src/modules/json.wren (hunk 11 above)
Comment (clsource) [reply src/modules/json.wren:126]: ok

File: src/modules/json.c (hunk 12 above)
Comment (clsource) [reply src/modules/json.c:125]: ok

File: src/vm.c (hunk 13 above)
Comment (clsource) [reply src/vm.c:302]: ok

File: docs/index.md [hunk 14]
```diff
@@ -17,7 +17,10 @@ Check out the documentation:
   * [input](modules/input)
   * [io](modules/io)
   * [math](modules/math)
+  * [json](modules/json)
```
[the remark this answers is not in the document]
clsource: ok

src/modules/json.wren
avivbeeri: Is it possible to support streaming, but also detect errors in the JSON formatting? If not, I think it's better that we disable the streaming option entirely, because silent failure like this isn't good for game development.

The user/developer might make a.mistake in their configuration, and the json module won't detect it at all, leading to avoidable confusion.

clsource: Let me check the lib for this option. And see what can be done 👍

src/modules/json.wren
```diff
@@ -0,0 +1,246 @@
...
+        substrings.add(JsonStream.escapechar(char, options))
+      }
+
+      // Compile error if you use normal escaping sequence
+      // so we have to use bytes to string method for the single " char
+      return String.fromByte(0x22) + substrings.join("") + String.fromByte(0x22)
+    }
+
+    if (value is List) {
+      var substrings = value.map { |item| Json.dumps(item) }
```
avivbeeri, line 204: `Json.dump` doesn't exist, I assume this was renamed to `Json.encode`.

src/modules/json.wren
```diff
@@ -0,0 +1,246 @@
...
+    var content = Json.encode(object, options)
+    FileSystem.save(path, content)
+    return content
+  }
+
+  static save(path, object) {
+    return Json.save(path, object, JsonOptions.ABORT_ON_ERROR)
+  }
+
+  // MARK: - General Manipulation
```
avivbeeri, line 176: What is this line for?

docs/modules/json.md
```diff
@@ -0,0 +1,71 @@
...
+### Constants
+
+#### `static NIL: Num`
+_Default_ value. No options selected.
+
+#### `static ESCAPE_SLASHES: Num`
+This will encode solidus character (`/`). When converting a `Map` object to a `String`. This encoding is optional and is useful when you need to embed `JSON` inside `HTML <script>` tags. By default _DOME_ does not escape slashes.
+
+#### `static ABORT_ON_ERROR: Num`
+By default _DOME_ does not _Abort_ when there is a _JSON parsing error_. Use this option when you need to trigger a `Fiber.abort()` on parse error.
```
avivbeeri, line 44: We've changed this behaviour to the default.

docs/modules/json.md
```diff
@@ -0,0 +1,71 @@
...
+### `static save(path: String, object: Object)`
+### `static save(path: String, object:Object, options:Num)`
+This will encode the object and then save the result to a file specified in `path`. With default or custom options.
+
+## JsonOptions
+
+### Constants
+
+#### `static NIL: Num`
+_Default_ value. No options selected.
```
avivbeeri, line 38: This isn't the default anymore.

docs/modules/json.md
```diff
@@ -0,0 +1,71 @@
...
+Use [Bitwise OR](https://wren.io/method-calls.html#operators) operator to select multiple options.
+
+```js
+Json.decode(myString, 
+	JsonOptions.ESCAPE_SLASHES |
+	JsonOptions.ABORT_ON_ERROR
+)
+```
+
+## JsonError
```
avivbeeri, line 57: It's not clear when this error is provided to the user, if at all.

src/modules/json.c
```diff
@@ -0,0 +1,131 @@
...
+  "\\" // A backslash.
+  "\a" // Alarm beep. (Who uses this?)
+  "\b" // Backspace.
+  "\f" // Formfeed.
+  "\n" // Newline.
+  "\r" // Carriage return.
+  "\t" // Tab.
+  "\v" // Vertical tab.
+  */
+  if(utf8cmp(value, "\0") == 0) {
```
avivbeeri, line 98: Tiny codestyle fix, put a space after the `if` here.

src/modules/json.c
```diff
@@ -0,0 +1,131 @@
...
+    result = "\\0";
+  } else if (utf8cmp(value, "\"") == 0) {
+    result = "\\\"";
+  } else if (utf8cmp(value, "\\") == 0) {
+    result = "\\\\";
+  } else if(utf8cmp(value, "\\") == 0) {
```
avivbeeri, line 104: Space after `if`

src/modules/json.c
```diff
@@ -0,0 +1,131 @@
...
+  } else if (utf8cmp(value, "\v") == 0) {
+    result = "\\v";
+  } else if (utf8cmp(value, "/") == 0) {
+    // Escape / (solidus, slash)
+    // https://stackoverflow.com/a/9735430
+    // The feature of the slash escape allows JSON to be embedded in HTML (as SGML) and XML.
+    // https://www.w3.org/TR/html4/appendix/notes.html#h-B.3.2
+    // This is optional escaping. Disabled by default.
+    // use JsonOptions.ESCAPE_SLASHES option to enable it
+    if((options & JSON_DOME_ESCAPE_SLASHES) != JSON_DOME_NIL_OPTIONS) {
```
avivbeeri, line 125: Space after `if`

avivbeeri:
Last few changes I've mentioned in comments above, and I think this PR is ready for merging! Thank you for your work on it.

src/modules/json.wren
clsource: In order to make the code easily readable, approachable, and manageable, a best practice to follow is that we divide its different aspects into groups and put a _// MARK:_ above each group with a descriptive comment.

https://nshipster.com/swift-documentation/#mark-todo-fixme

For now is just a simple comment, but maybe in the future would be handy for code analyzing tools.

docs/modules/json.md
clsource: It would be provided if the _ABORT_ON_ERROR_ option is disabled.

```js
static decode(value, options) {
    var stream = JsonStream.new(value, options)
    stream.begin()

    var result = stream.result
    if (stream.error.found) {
      result = stream.error
    }

    stream.end()
    return result
  }
```

avivbeeri: I know that, but it's not stated in the documentation, so other people using these classes won't be able to tell.

clsource: added: 

This object is returned on calls to _Json.decode_ only if the  `JsonOptions.ABORT_ON_ERROR` default behaviour is disabled and a parse error was found.

to the JsonError docs

src/modules/json.wren
avivbeeri: For the moment, this form of notation (as well as other formal documentation comments) aren't in keeping with the existing DOME code style. 

Good code commenting is a best practice across all projects, but it's up to individual projects how that practice is implemented, and to do that here is out of scope for this PR.

I don't mind methods being grouped in related blocks, and indicated as such, I just don't want to commit to a specific tool/system/syntax at this time.

Please remove them for the time being.

clsource: ok

PR #85, opened 2020-09-27
[WIP] added unit testing library and sample tests

clsource:
For unit testing different components in DOME.

imagen

```js
import "math" for Math
import "unit" for Assert

class MathTests {
  static name { "modules/math.wren" }

  static thatSinWorks {[
      "that `sin` function works",
      Fiber.new {
          Assert.equal(Math.sin(2), 2.sin)
      }
  ]}

  static thatCosWorks {[
      "that `cos` function works",
      Fiber.new {
          Assert.equal(Math.cos(2), 2.cos)
      }
  ]}

  static thatTanWorks {[
      "that `tan` function works",
      Fiber.new {
          Assert.equal(Math.tan(2), 2.tan)
      }
  ]}

  static all {[
    thatSinWorks,
    thatCosWorks,
    thatTanWorks
  ]}
}

```

```js
import "unit" for Runner
import "./math.test" for MathTests
import "./vector.test" for VectorTests

class Game {
    static init() {
        // Add your tests here
        Runner.run(MathTests)
        Runner.run(VectorTests)

        Runner.end()
    }
}
```


avivbeeri:
I don't find TDD or unit testing particularly helpful during game development, so its primary purpose would be for testing DOME specifically. In that instance, it'd be better not to ship unit testing tools to every user just so the project can do tests on itself. They should be able to include them optionally, if they need them.

clsource:
I think DOME can also be used for making multimedia desktop applications or helper tools for game development (Maybe a map builder for a specific game?).

But it's true that DOME's primary focus is games. 
Maybe this file should be kept in _domepunk_ and stay on the dome engine codebase as an aid for testing the engine internals. 

This was a proof of concept anyway 👍

avivbeeri:
I do agree that DOME could be used for other applications (although mostly I focus it towards games, and game development tools), I think keeping it external, as a general tool for all Wren projects is the right approach.

Issue #87, opened 2020-09-28
[idea] Support command line params in Wren


clsource:
Dome could provide some API to access command line params.
So games can change their behaviours depending on these params.

Some quick draft

```js

import "dome" for Args
class Game {
    static init() {
           Args.dome // A Map that stores the argv with DOME specific params (like --debug)
          Args.app // A Map that stores the argv with all the other Not standard DOME params
         Args.all // Args.dome + Args.app params
    }
}
```

avivbeeri:
In general, the order of arguments is significant in command line tools, so
an array-style of access may be better.

clsource:
One thing that I would like to know if this lib should be on its own namespaces like

```js
import "args" for Args
```

or its best to add to an available module like `dome`.

Its clearly that this module would need access to the C backend the same as the `Version` class.

avivbeeri:
Args processing should be considered part of the dome.Process class, as
it's a property of the running process.

PR #89, opened 2020-10-01
added arguments for wren

clsource:
Following #87 I made a small test concept.
Figure it out it was simpler to just accept a string for the user.
The user would be responsible to parse and use this string as needed.

```js
import "dome" for Process

class Game {
    static init() {
      System.print(Process.args)
    }
    static update() {}
    static draw(dt) {
    }
}
```

imagen

```c
internal void
PROCESS_args(WrenVM* vm) {
  char * args = engine->args;
  if(args) {
    wrenSetSlotString(vm, 0, args);
    return;
  }
  wrenSetSlotNull(vm, 0);
}
```


avivbeeri:
I appreciate the idea, but I think we can provide arguments better. DOME
uses skeeto/optparse.h to process arguments. We just need to capture the
ones which aren't used by dome.

Maybe we can use a `--` to seperate engine and usercode arguments, and then
provide an array of arguments to the user.

`./dome -d --record=data.gif main.wren -- --edit=tracks/1.dat --new`

As an example, would then supply to process.args: `[ "--edit=tracks/1.dat",
"--new" ]`

Not a simple implementation, due to memory allocation, but more useful.

avivbeeri:
I appreciate the idea, but I think we can provide arguments better. DOME
uses skeeto/optparse.h to process arguments. We just need to capture the
ones which aren't used by dome.

Maybe we can use a `--` to seperate engine and usercode arguments, and then
provide an array of arguments to the user.

`./dome -d --record=data.gif main.wren -- --edit=tracks/1.dat --new`

As an example, would then supply to process.args: `[ "--edit=tracks/1.dat",
"--new" ]`

Not a simple implementation, due to memory allocation, but more useful.

clsource:
Well for the Process class it can be implemented with a simple string separation.

- `args`: The raw string value
- `arguments`: The string value converted to a list separating by spaces.

```js
class Process {
  foreign static f_exit(n)
  foreign static args
  
  static arguments {
    if (Process.args) {
      return Process.args.split(" ")
    }
    return []
  }
  
  static exit(n) {
    f_exit(n)
    Fiber.suspend()
  }

  static exit() {
    exit(0)
  }
}
```

So my idea is to catch anything that is not recognized by DOME and store it in the `engine.args` character array and return it to a string.

clsource:
We could also use some similar strategy to the wren-cli

https://github.com/wren-lang/wren-cli/blob/main/src/module/os.c

clsource:
imagen

src/main.c
```diff
@@ -209,12 +209,17 @@ int main(int argc, char* args[])
     {"version", 'v', OPTPARSE_NONE},
     {"record", 'r', OPTPARSE_OPTIONAL},
     {"scale", 's', OPTPARSE_REQUIRED},
+    {"args", 'a', OPTPARSE_OPTIONAL}, // Game exclusive settings
```
avivbeeri, line 212: We probably don't need this if we can provide all the command line flags to the game

avivbeeri:
There's a data.gif which is part of this PR, and that seems unnecessary to me.

src/modules/dome.c
```diff
@@ -8,6 +8,20 @@ PROCESS_exit(WrenVM* vm) {
...
 }
 
+internal void
+PROCESS_args(WrenVM* vm) {
+  ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
+  
+  wrenEnsureSlots(vm, 2);
+  wrenSetSlotNewList(vm, 0);
+
+  for (int i = 0; i < engine->argc; i++)
```
avivbeeri, line 18: If this code-block was taken from the Wren-CLI project, we should make some note of that fact here, for licensing reasons.

src/modules/dome.wren
```diff
@@ -45,6 +45,10 @@ class Version {
 
 class Process {
   foreign static f_exit(n)
+  foreign static args
```
avivbeeri, line 48: This method should be `f_args`, since it's intended to be an internal/private property, rather than a user-facing one.

avivbeeri:
In general, I like the approach here. We just need some documentation and a couple other points I raised and we are good to go:

- [ ] Documenting the Process.arguments API
- [ ] Add comments about Wren licensing.
- [ ] Rename Process.args to Process.f_args
- [ ] Remove data.gif
- [ ] Remove the `-a` option in the optparse configuration as it is now unnecessary.

avivbeeri (changes requested):
In general, I like the approach here. We just need some documentation and a couple other points I raised and we are good to go:

- [ ] Documenting the Process.arguments API
- [ ] Add comments about Wren licensing.
- [ ] Rename Process.args to Process.f_args
- [ ] Remove data.gif
- [ ] Remove the `-a` option in the optparse configuration as it is now unnecessary.

src/modules/dome.wren
clsource: I think both should be user facing since the idea is:

- args: Returns all arguments.
- arguments: Returns user only arguments ( args - DOME arguments - `./dome` )

avivbeeri: We might need to check if it does return all arguments. I think optparse might omit `--` if it's an argument on it's own.

src/modules/dome.wren
```diff
@@ -45,6 +45,10 @@ class Version {
...
+
+  static arguments { args[2..-1] }
```
avivbeeri, line 50: This actually just crashes if you haven't got enough arguments.

There's really no way to know how many DOME-specific arguments are going to be used in a commandline call.

src/modules/dome.wren
avivbeeri: Having double-checked, what happens is that once `--` is encountered, everything after it is assumed not to be commandline options. This can lead to some weird behaviour if you haven't manually specified as the first non-option argument to be your main wren file. Tricky problem to resolve without changing the commandline usage slightly.

clsource: I think is fine for now rename `args` to `f_args` since it will contain information normally not relevant to the final user.

and `arguments` would contain everything except `./dome`

if you pass a string to the arguments.  it would return the proper value if there is a match.

```js
static arguments { Process.f_args[1..-1] }

static arguments(needle) {
     var result = null
    Process.f_args.each { |item| 
          if (item.contains(needle)) {
                result = item
         }
   }
  return result
}
```

src/modules/dome.wren
clsource: The a little improvement should be done for checking `f_args.count` before deleting the first element

clsource:
I found out that if you include the filename and omit using `--` you can have the functionality

```js
class Process {
  foreign static f_exit(n)
  foreign static f_args

  static arguments {
    var args = Process.f_args
    if (args.count > 1) {
      return args[2..-1]
    }
    return args
  }

  static arguments(needle) {
    var args = Process.arguments
    var result = null
    Fiber.new {
      args.each {|value|
        if ("%(value)".contains(needle)) {
          result = value
          // Skip all other values
          Fiber.abort()
        }
      }
    }.try()
    return result
  }
}
```

```
import "dome" for Process

class Game {
    static init() {
      System.print(Process.arguments)
      System.print(Process.arguments("edit"))
    }
    static update() {}
    static draw(dt) {
    }
}
```
imagen

clsource:
* [ ]  Documenting the Process.arguments API
* [x] Add comments about Wren licensing.
* [x] Rename Process.args to Process.f_args
* [x] Remove data.gif
* [x] Remove the -a option in the optparse configuration as it is now unnecessary.

src/modules/dome.wren
```diff
@@ -45,6 +45,24 @@ class Version {
 
 class Process {
   foreign static f_exit(n)
+  foreign static f_args
+
+  static arguments {Process.f_args}
+  static arguments(needle) {
+    var args = Process.arguments
+    var result = null
+    Fiber.new {
```
avivbeeri, line 54: I don't think we need a whole fiber just to abort a loop early. You can do:

```
for (value in args) {
  if (value.contains(needle)) {
    result = value
  }
}
```
However, I have concerns for two reasons: 
1) any argument could "contain" the needle and it would be caught here. You probably want `value.startsWith("--%(needle)")` for the specific effect you are trying to achieve.

2) this doesn't support "short" options: `--edit` could be just `-e` depending on what you wanted to enable. This is something optparse allows.

Command line processing has a lot of edge cases.

avivbeeri (changes requested):
I'm going to be quite firm here. I do not want DOME to pre-process command line arguments, except to provide an array of them as strings. All the parsing and interpretation should be left to the user code.

src/modules/dome.wren
clsource: I think for the sake of minimalism. Is fine for DOME just returning the whole array of arguments. And leave the parsing to the final user 👍

clsource: The `--` is something that could be ditched since it conflicts with optparse

the convention could be that everything after the filename is part of the args 💯 

`./dome main.wren edit=tracks/1.dat new`

avivbeeri: The optparse library is what handles the `--`, so we can't easily ditch it. I was asking in the discord whether most people actually specify a particular script name, because we could easily add a named option for that, and it would fix the behaviour to be consistent with what optparse expects.

clsource: Well by ditching `--` I mean only for the custom arguments.

Example: `./dome -d --record=data.gif main.wren edit=tracks/1.dat new`
This would be read as `[./dome, -d, --record=data.gif, main.wren, edit=tracks/1.dat, new]`

So you can say say ever argument that starts with `-` or `--` is a DOME argument. And everything without (after the filename) it can be considered a custom argument.

avivbeeri: Sure, although we can easily say: "anything without `-`, `--` and anything following a lone `--` are arguments available for usercode to use". 

But for any of this to work reliably, we need to change how DOME selects it's main file, so that it's a named argument. It would need to be something like `./dome -f main.wren ...` or `./dome --file=main.wren ...`, although the name/shortening are up for debate. The alternative is to swap out `optparse` for a different mechanism of parsing command line arguments entirely.

avivbeeri:
What's the state of this PR?

clsource:
It should be defined the proper way of dealing with custom arguments (non dome)
Currently it supports reading all the arguments as is.

I think `main.wren` should be considered in similar fashion to `./dome` 
By convention the first argument (that ends with *.wren) should be the filename.

avivbeeri:
The path given to DOME doesn't explicitly have to be a file, though. You
can point it to a directory, and it will automatically detect a game.egg,
or main.wren file if available.

avivbeeri:
So, I'd expect the "usage" to look like this:

```
Usage:
  dome [-c | --console] [-d | --debug] [-r<gif> | --record=<gif>] [-b<buf> | --buffer=<buf>] [entry path] [--] [non-DOME arguments]
  dome -h | --help
  dome -v | --version

Options:
  -b --buffer=<buf>   Set the audio buffer size (default: 11)
  -c --console        On Windows, this opens a log terminal for debugging.
  -d --debug          Enables debug mode.
  -h --help           Show this screen.
  -v --version        Show version.
  -r --record=<gif>   Record video to <gif>.
```

Technically the entry path and non-DOME arguments would be grouped together. How can we handle things during the valid case where an entry path isn't provided, but other arguments are?

avivbeeri:
After some experimenting, I've settled on an approach for handling arguments I am happy with, and have opened a PR for it.

Please can you give it a look? https://github.com/avivbeeri/dome/pull/109

avivbeeri:
This was successfully implemented in #109 so I am closing this PR.

Issue #91, opened 2020-10-06
Add which version of Wren is in use to the Version message


avivbeeri:
It makes sense to list the versions of all our core dependancies when the version string is requested, just in case it's relevant.

(For example, certain language API/features may only be available in newer Wren implementations)

PR #95, opened 2020-10-07
Add Dependencies Handler

clsource:
Addresses #91


avivbeeri:
This is interesting, but not what I meant for this issue. I intended this to be a change to `printUsage` in `src/main.c`, which outputs the current DOME and SDL version to the command line. I wanted the Wren version to be listed here also.

PR #97, opened 2020-10-09
Allow lines of arbitrary thickness to be drawn

### LINKED ISSUE #92: 2D drawing - Thicker lines
This isn't an issue as such but I thought you might be interested to know that I've been using DOME to successfully create some fancy 2D static images. Basically, I just do all the drawing from the init() method or from methods called by it. I leave the update() and draw() methods empty.

It might be worth mentioning in your blurb that this is possible as it may widen the project's appeal.

One thing that would be useful for future versions is to be able to draw lines etc. with a different thickness than one pixel. At present, I just draw the line the required number of times after offsetting the start and end co-ordinates appropriately.

avivbeeri:
This would close #92 by adding an overloaded `Canvas.line` option to include line thickness.


src/engine.c
```diff
@@ -524,7 +551,8 @@ ENGINE_line_high(ENGINE* engine, int64_t x1, int64_t y1, int64_t x2, int64_t y2,
   int64_t y = y1;
   int64_t x = x1;
   while(y <= y2) {
-    ENGINE_pset(engine, x, y, c);
+    ENGINE_drawMask(engine, x, y);
+    // ENGINE_pset(engine, x, y, c);
```
avivbeeri, line 555: Need to clean up the commented out `pset` calls.

src/engine.c
```diff
@@ -510,6 +510,33 @@ ENGINE_blitLine(ENGINE* engine, int64_t x, int64_t y, int64_t w, uint32_t* buf)
   memcpy(line, bufStart, lineWidth * 4);
 }
 
+internal uint32_t*
+ENGINE_createMask(ENGINE* engine, uint64_t originalSize, uint32_t color) {
+  uint64_t size = originalSize;
+  if (size % 2 == 0) {
+    size = size + 1;
```
avivbeeri, line 517: This results in odd-thickness only lines, centered on the given coordinates.

Issue #98, opened 2020-10-10
Reusable third-party module support


avivbeeri:
The goal of this issue is to discuss the implementation of something like NodeJS or Python modules, which can be developed and distributed for use across multiple projects with ease.

Any changes to the module-loading system should be backwards compatible with current projects, otherwise we will have to defer it to DOME 2.0.0

I think my own preference would be to extend the `.egg/.tar` loading system and use that file format for distributing modules.

As far as indicating that part of a module is being imported, I'd expect something like:
```
import "@domepunk" for PRNG
```
or, possibly allowing access to sub-files within a module using:
```
import "@domepunk/sub/path/to/Random" for PRNG
```

The hard bit is that DOME would need to keep track of which "module context" it was in, in order to resolve modules properly, as well as other rabbit-hole issues.

clsource:
# Dome Module Imports

For simplifying code reutilization and modularization, 
this is a proposed way for _DOME_ to handle dependencies.

## modules.wren

This is a simple wren file that contains
each module path. Is inspired on https://github.com/WICG/import-maps

```js
var Imports = {
    "domepunk": "modules/domepunk/dp.wren",
    "domepunk/": "modules/domepunk", // allow imports for files inside the domepunk/ dir
}
```

On startup _DOME_ will load this file (if exists) and store the paths for reference later
when loading `imports` inside wren files.

The filename could be overrited using `--importmap=<FILE>` CLI flag.

### tar files

Paths that end with _tar_ will be considered as a _tar_ file.
Using `@` after the file extension it would mean a file inside the _tar_.

-
```js
var Imports = {
    "domepunk": "modules/domepunk.tar",
    "domepunk/math": "modules/domepunk.tar@science/math"
}
```

### importing

- Using `@` at the start of the import will use the _importmap_ path first.

- Without using `@` it would follow the traditional way of importing and use the _importmap_ last for resolve it.

- If the paths begins with `/` (_"/domepunk"_) it would consider the base path of the _dome_ executable.

- If the path does not begins with `/` (_"domepunk"_) it would consider the base path of the _main.wren_ file.

- If the path begins with `./` (_"./domepunk"_) it would consider the base path of the current file begin executed.

```js
// Will use the math module path defined in modules.wren
import "@math" for Math

// Will use the same path for the _dome_ executable as the base path. if not found will look into the modules.wren path
import "/math" for Math

[...]

PR #99, opened 2020-10-11
added proof of concept for module imports

clsource:
This is just a simple prototype for testing the concept of the `importmap` discussed in #98 

- `dome.ini` file stores the mapping for the imports.
```ini
# Dome Configuration
[imports]
code = example/
code2 = example/
```
Usage

```js
import "code" for Code // maps to example/code.wren
import "code2" for Code2 //maps to example/code2.wren
```


avivbeeri:
How does this behave if those submodules want to depend on their own dependancies?

clsource:
```js

import "code" for Code

class Code {
    static init() {
        System.print("Code 1 Called")
        Code2.init()
    }
}

```

Reading from filesystem: /Users/clsource/Code/dome/dome/example/code.wren
code:4: Error at 'Code': Module variable is already defined.
code:7: Error at 'Code2': Variable is used but not defined.
Runtime error: Could not compile module 'code'.
  2: main

clsource:
This is just a simple map of "key = path".
were `key` is the import value.

For a more complex scenario like:

```
import "code/code2" for Code2
```

It would be needed more rules for parsing the import string. 
For example detects if it contains `/`, `@` and other rules.
And then map the corresponding key to the final path to look for the file.

clsource:
If submodules would like to depend on other submodules I think.
The least complex way of solving is having the single source of truth of the root _dome.ini_.

Submodules would require the final user to define the values inside _dome.ini_ which they will use inside their codebase.

Example:

_DomePunk_  would require that the path _domepunk/_ is defined in _dome.ini_

Then the game in  _main.wren_ Would use:

```js
import "domepunk/dp" for Dp
```

In _domepunk/dp.wren_ Would use

```js
import "domepunk/utils/clock" for Clock
```

And _dome.ini_ would be defined as

```ini
[imports]
dumpunk = modules/domepunk/dp
domepunk/ = modules/domepunk
```

So everytime when `import "domepunk" for Dp` will be mapped as: `modules/domepunk/dp.wren`
And when `import "domepunk/utils/clock" for Clock` is used inside _dp.wren_ it would be mapped to: 
`modules/domepunk/utils/clock.wren`

clsource:
Now the module loader supports using `solidus` to define a namespace.

```c
internal WrenForeignMethodFn VM_bind_foreign_method(
    WrenVM* vm,
    const char* module,
    const char* className,
    bool isStatic,
    const char* signature) {

  // This file is seperate because it has a Copyright notice with it.
#include "signature.c.inc"

  MAP moduleMap = engine->moduleMap;
  return MAP_getFunction(&moduleMap, module, fullName);
}

internal char* VM_load_module(WrenVM* vm, const char* name) {
  MAP moduleMap = engine->moduleMap;

  if (DEBUG_MODE) {
    ENGINE_printLog(engine, "Loading module %s from ", name);
  }

  // Check against wren optional modules
#if WREN_OPT_META
  if (strcmp(name, "meta") == 0) {
    if (DEBUG_MODE) {
      ENGINE_printLog(engine, "wren\n", name);
    }
    return NULL;
  }
#endif
#if WREN_OPT_RANDOM
  if (strcmp(name, "random") == 0) {
    if (DEBUG_MODE) {
      ENGINE_printLog(engine, "wren\n", name);
    }
    return NULL;
  }
#endif

  // Check against dome modules
  char* module = (char*)MAP_getSource(&moduleMap, name);

  if (module != NULL) {
    if (DEBUG_MODE) {
      ENGINE_printLog(engine, "dome\n", name);
    }
    return module;
  }

  // Otherwise, search on filesystem
  char* extension = ".wren";
  char* path;
  path = malloc(strlen(name)+strlen(extension)+1); /* make space for the new string (should check the return value ...) */
  strcpy(path, name); /* add the extension */
  strcat(path, extension); /* add the extension */

  if (DEBUG_MODE) {
    ENGINE_printLog(engine, "%s\n", engine->tar ? "egg bundle" : "filesystem");
  }

  // This pointer becomes owned by the WrenVM and freed later.
  char* file = ENGINE_readFile(engine, path, NULL);
  if (file == NULL) {
    
    if (DEBUG_MODE) {
      ENGINE_printLog(engine, "Path not found. Looking in dome.ini\n", path);
    }

    // Search for / in the name
    char * solidus = "/";
    char * full = strdup(name);
    
    #define MAX_PATHS 255

    char paths[MAX_PATHS][MAX_PATHS];
    int index = 0;
    
    char * token = strtok(name, solidus);
    while (token != NULL && index < MAX_PATHS) {
      strcpy(paths[index], token);
      token = strtok(NULL, solidus);
      index++;
    }

    // Search for the import key
    char * key = strdup(full);

    // There are more values after solidus
    // We append the solidus for the key
    int isSolidus = 0;
    if (index > 1) {
      strcpy(key, paths[0]);
      strcat(key, solidus);
      isSolidus = 1;
    }

    // Look for the modules inside config
    char * importPath = CONFIG_readValue(&engine->config, "imports", key);
    if (importPath != NULL) {
      path = malloc(strlen(importPath) + strlen(full) + strlen(extension) + 1);
      strcpy(path, importPath);

      if (isSolidus == 1) {
        // We dont need to append the part with solidus just the next one
        for(int i = 1; i < index; i++) {
          strcat(path, paths[i]);
          if (i < index - 1) {
            strcat(path, solidus);
          }
        }
      } else {
        strcat(path, full);
      }

      strcat(path, extension);
      file = ENGINE_readFile(engine, path, NULL);

    } else {
      ENGINE_printLog(engine, "No import Path Found for Module %s\n", name);
    }
  }

  free(path);
  return file;
}
```

clsource:
Example Log

```
Reading from filesystem: /Users/clsource/Code/dome/dome/dome.ini
Reading from filesystem: /Users/clsource/Code/dome/dome/main.wren
Loading module graphics from dome
Loading module vector from dome
Loading module image from dome
Loading module font from dome
Loading module io from dome
Loading module code from filesystem
Reading from filesystem: /Users/clsource/Code/dome/dome/code.wren
Path not found. Looking in dome.ini
Reading from filesystem: /Users/clsource/Code/dome/dome/example/code.wren
Loading module code/code2 from filesystem
Reading from filesystem: /Users/clsource/Code/dome/dome/code/code2.wren
Path not found. Looking in dome.ini
Reading from filesystem: /Users/clsource/Code/dome/dome/example/code2.wren
Code 1 Called
Code 2 Called

```

PR #104, opened 2020-10-21
Added createDirectory functionality

### LINKED ISSUE #102: Allow DOME to create folders
Proposed API:
`FileSystem.createDirectory(path)` would behave like `mkdir -p` on unix systems, where any missing folders are created on the way.

It should Fiber.abort if the directory couldn't be created for some reason.

siddhantrao23:
Simulates `mkdir -p` and aborts if directory couldn't be created. Testing done on linux (Ubuntu 18.04). Windows tests are needed.

Fixes #102 

Uses code from [mkdirp](https://github.com/stephenmathieson/mkdirp.c)

Example usage:
```
import "io" for FileSystem
FileSystem.createDirectory("test/me/out")
```
The permissions given to the directory are hard coded to `0777` as of now which should be changed if required.

Do let me know of any changes!

### DIFF
--- a/src/include/mkdirp/mkdirp.c
+++ b/src/include/mkdirp/mkdirp.c
@@ -1,0 +1,70 @@
+#include <unistd.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "mkdirp.h"
+
+#ifdef _WIN32
+#define PATH_SEPARATOR   '\\'
+#else
+#define PATH_SEPARATOR   '/'
+#endif
+
+char *
+path_normalize(const char *path) {
+    if(!path) return NULL;
+
+    char *copy = strdup(path);
+    if(copy == NULL) return NULL;
+    char *ptr = copy;
+
+    for (int i = 0; copy[i]; ++i) {
+        *ptr++ = path[i];
+        if (path[i] == '/') {
+            ++i;
+            while(path[i] == '/') {
+                --i;
+            }
+        }
+    }
+    *ptr = '\0';
+    return copy;
+}
+
+int
+mkdirp(const char *path, mode_t mode) {
+    char *pathname = NULL;
+    char *parent = NULL;
+
+    if(path == NULL) return -1;
+    pathname = path_normalize(path);
+    if (pathname == NULL) goto fail;
+
+    parent = strdup(pathname);
+    if (parent == NULL) goto fail;
+
+    char *p = parent + strlen(parent);
+    while(PATH_SEPARATOR != *p && p != parent) {
+        --p;
+    }
+    *p = '\0';
+
+    if(p != parent && mkdir(parent, mode != 0)) goto fail;
+    free(parent);
+
+#ifdef _WIN32
+    int rc = mkdir(pathname);
+#else
+    int rc = mkdir(pathname, mode);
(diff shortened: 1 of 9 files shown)


src/include/mkdirp/mkdirp.c
```diff
@@ -0,0 +1,82 @@
+#include <unistd.h>
```
avivbeeri, line 1: We should have a full copy of the MIT license from the mkdirp library in here, since we are copying and redistributing this library.

src/include/mkdirp/mkdirp.h
```diff
@@ -0,0 +1,21 @@
+
+//
+// mkdirp.h
+//
+// Copyright (c) 2013 Stephen Mathieson
+// MIT licensed
```
avivbeeri, line 6: We should have a full copy of the MIT license from the mkdirp library in here, since we are copying and redistributing that code.

src/modules/io.c
```diff
@@ -308,3 +308,25 @@ internal void
...
+  const char* fullPath;
+  mode_t mode = 0777;
+  if (path[0] != '/') {
+    char* base = BASEPATH_get();
+    fullPath = malloc(strlen(base)+strlen(path)+1);
+    strcpy((char*)fullPath, base);
+    strcat((char*)fullPath, path);
+  } else {
+    fullPath = path;
+  }
```
avivbeeri, line 324: I realise that this is my own fault, but it would be good to pull the path-resolving logic out into its own method, so that we aren't copy-pasting it between functions. I'd suggest putting this method in `src/io.c`, as:

```c
char* resolvePath(char* partialPath, bool* shouldFree) {
    const char* fullPath;
    mode_t mode = 0777;
    if (partialPath[0] != '/') {
      char* base = BASEPATH_get();
      fullPath = malloc(strlen(base)+strlen(partialPath)+1);
      strcpy((char*)fullPath, base);
      strcat((char*)fullPath, partialPath);
      if (shouldFree != NULL) {
        *shouldFree = true;
      }
    } else {
      fullPath = partialPath;
      if (shouldFree != NULL) {
        *shouldFree = false;
      }
    }

    return fullPath;
}
```

Then we can call it as 
```c
bool shouldFree;
char* fullPath = resolvePath(path, &shouldFree);
// Use the full path for something here.
// For example:
// int result = mkdirp(fullPath, mode);
if (shouldFree) {
  free((void*)fullPath);
}
```

src/modules/io.wren
```diff
@@ -13,6 +13,7 @@ class FileSystem {
   foreign static save(path, buffer)
   foreign static prefPath(org, app)
   foreign static basePath()
+  foreign static createDirectory(path)
```
avivbeeri, line 16: This needs to be added to the documentation. The correct file for that is `docs/modules/io.md`.

avivbeeri (changes requested):
The general implementation looks fine. I've suggested a few changes to handle our licensing responsibilities, and a small refactor to improve DOME's overall codebase.

avivbeeri (approved):

avivbeeri:
Great work, thanks so much!

siddhantrao23:
I had a lot of fun working on it. Super cool project!!

PR #105, opened 2020-10-22
Adding basic github actions

avivbeeri:
When this should run is still up for debate.
At minimum, when a tag is created in the repository, but is there value in other builds?


Makefile
```diff
@@ -84,7 +84,9 @@ ifneq (, $(findstring MINGW, $(SYS)))
 endif
 
 ifneq (, $(findstring Linux, $(SYS)))
-  CFLAGS += -Wno-discarded-qualifiers -Wno-clobbered --no-pie
+  CFLAGS += -Wno-discarded-qualifiers -Wno-clobbered 
+	LDFLAGS += -ldl
```
avivbeeri, line 88: Check if we still need to add this manually.

PR #109, opened 2020-10-25
Process Arguments

### LINKED ISSUE #87: [idea] Support command line params in Wren

avivbeeri:
This is my take on #87 and #89 , inspired by how NodeJS handles its commandline parameters.

The resulting usage now looks like this:
```
Usage:
  dome [options]
  dome [options] [--] entry_path [arguments]
  dome -h | --help
  dome -v | --version
```
The first case, where no arguments are provided, will follow the established auto-resolution procedure:
1) Is there `game.egg` in the current working directory? if so, execute it
2) Is there `main.wren` in the current working directory? if so, execute it.
3) Print an error, since no entry point could be found.

For the second case, where an entry path (the first non-option argument) and some number of arguments are provided, will do the following:
1) Is the entry path a directory? If so, follow the auto-resolution as described above, using that directory.
2) Else, use the entry path:
2a) Test whether it is an .egg file, and use it if it is.
2b) Otherwise, treat it as a .wren file, and process accordingly.

`Process.args` will return all non-option arguments used to invoke DOME, but it will always include the invoked executable path and the full path of the entry point (by using the basepath).

(Closes #87 and Closes #89 )

### DIFF
--- a/docs/modules/dome.md
+++ b/docs/modules/dome.md
@@ -13,8 +13,10 @@ It contains the following classes:
 ## Process
 
 ### Static Fields
-#### `static arguments: String[]`
-Returns a string list of all command line arguments used to invoke DOME's execution, in the other they were provided.
+#### `static args: String[]`
+Returns a string list of all non-option command line arguments used to invoke DOME's execution. The first two elements of the returned list will be:
+1) the path and name of DOME's invokation.
+2) The entry path, combined with the evaluated basepath.
 
 ### Static Methods
 
--- a/docs/modules/dome.md
+++ b/docs/modules/dome.md
@@ -12,6 +12,10 @@ It contains the following classes:
 
 ## Process
 
+### Static Fields
+#### `static arguments: String[]`
+Returns a string list of all command line arguments used to invoke DOME's execution, in the other they were provided.
+
 ### Static Methods
 
 #### `static exit(): Void`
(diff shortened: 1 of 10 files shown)


clsource:
Did some tests.
Seems to be working fine.

```js
import "dome" for Process

class Game {

    static init() {
        System.print(Process.args)
        Process.exit()
    }
}
```

```sh
./dome --debug main.args.wren arg1 arg2 arg3
Reading from filesystem: /Users/clsource/Code/dome/dome/main.args.wren
Loading module dome from dome
[./dome, /Users/clsource/Code/dome/dome/main.args.wren, arg1, arg2, arg3]
```

```sh
./dome --debug main.args.wren -- arg1 arg2 arg3
Reading from filesystem: /Users/clsource/Code/dome/dome/main.args.wren
Loading module dome from dome
[./dome, /Users/clsource/Code/dome/dome/main.args.wren, arg1, arg2, arg3]
```

clsource (approved):
Seems fine

avivbeeri:
Thank you for giving this a review, as well as your assistance in designing this feature :)

PR #111, opened 2020-10-26
Mouse Position and Scroll vectors

avivbeeri:
I find myself often wanting Mouse co-ordinates as a vector, so this is a convenience method for supplying that.


avivbeeri:
I think I want to get Mouse scroll options in here as well, just to round out the Mouse functionality.

PR #117, opened 2020-11-13
added missing encoding use case

clsource:
When passing a custom class the encoder would fail
since it would return null.

This invokes the `toString` method in the object as the default encoding.

### DIFF
--- a/src/modules/json.wren
+++ b/src/modules/json.wren
@@ -184,6 +184,9 @@ class Json {
       }
       return "{" + substrings.join(",") + "}"
     }
+
+    // Default behaviour is to invoke the toString method
+    return value.toString
   }
 
   static encode(value) {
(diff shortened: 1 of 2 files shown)


avivbeeri:
I'm not 100% certain this is the best behaviour. Half makes me wish we had a form of "serialisable" interface first.

clsource:
Its a Wren standard using `toString` to convert any object to a serializable string.
so if someone wants to serialize a complex object, it just needed to override the `toString` method.

```js
import "json" for Json

class MyClass {
  toString {Json.encode(toMap)}
  toMap {{
  "property1": myproperty
 }}

 myproperty {_property}

 construct new(property) {
  _property = property
 }
}

var obj = MyClass.new(true)
System.print(Json.encode(obj))
```

src/modules/json.wren
```diff
@@ -184,6 +184,9 @@ class Json {
       }
       return "{" + substrings.join(",") + "}"
     }
+
+    // Default behaviour is to invoke the toString method
```
avivbeeri, line 188: We should document this for users of the `Json` class.

avivbeeri (changes requested):

avivbeeri (approved):

PR #132, opened 2020-12-11
[Exploration] Replacing Random implementation with noise-based RNG

avivbeeri:
Based on the Noise function Squirrel3 provided in [this talk](https://www.youtube.com/watch?v=LWFzPP8ZbdU), This is a drop-in replacement for the built-in Random module.

As a bonus, this could be used for other noise-based techniques, such as procedural generation of levels and similar.

I need to run some benchmarks, but it's an interesting feature to provide. Normally, I'd want this kind of thing to be provided externally, but because it can wholesale replace the built-in module, it makes sense to include it.

We also add a `platform` module and `Platform` class (also accessible via the `dome` module) to fetch system time, for seeding the PRNG.


avivbeeri:
Made some adjustments by moving the Noise implementation and `Random.float` methods into the C, and according to benchmarks, there is a speed improvement compared to the built-in Wren module, although it is slight.

Statistic testing still needs to be done to decide if it's reasonable to swap the random number generator out for our implementation.

PR #151, opened 2021-01-30
Implemented wren2c command

clsource:
Simple command that helps plugin development by transforming _Wren_ code to _C_ strings.

Usage: `./dome --wren2c <input> <module name>? <output>?`

- input: File to read.
- module name: Optional. If not present will use a random string.
- output: Optional. If not present will use the input + .inc extension. (E.g. `main.wren.inc`)

Example: 

- `./dome --wren2c main.wren`
- `./dome -w main.wren` (shorthand)

Output:

```c
// Generated automatically using ./dome --wren2c main.wren Do not modify
const char wrenModule_oFWKSfTl2n[196] = {'i', 'm', 'p', 'o', 'r', 't', ' ', '"', 'g', 'r', 'a', 'p', 'h', 'i', 'c', 's', '"', ' ', 'f', 'o', 'r', ' ', 'C', 'a', 'n', 'v', 'a', 's', ',', ' ', 'C', 'o', 'l', 'o', 'r', '\n',
'c', 'l', 'a', 's', 's', ' ', 'G', 'a', 'm', 'e', ' ', '{', '\n',
' ', ' ', ' ', ' ', 's', 't', 'a', 't', 'i', 'c', ' ', 'i', 'n', 'i', 't', '(', ')', ' ', '{', '}', '\n',
' ', ' ', ' ', ' ', 's', 't', 'a', 't', 'i', 'c', ' ', 'u', 'p', 'd', 'a', 't', 'e', '(', ')', ' ', '{', '}', '\n',
' ', ' ', ' ', ' ', 's', 't', 'a', 't', 'i', 'c', ' ', 'd', 'r', 'a', 'w', '(', 'd', 't', ')', ' ', '{', '\n',
' ', ' ', ' ', ' ', ' ', ' ', 'C', 'a', 'n', 'v', 'a', 's', '.', 'p', 'r', 'i', 'n', 't', '(', '"', 'D', 'O', 'M', 'E', ' ', 'I', 'n', 's', 't', 'a', 'l', 'l', 'e', 'd', ' ', 'S', 'u', 'c', 'c', 'e', 's', 's', 'f', 'u', 'l', 'l', 'y', '.', '"', ',', ' ', '1', '0', ',', ' ', '1', '0', ',', ' ', 'C', 'o', 'l', 'o', 'r', '.', 'w', 'h', 'i', 't', 'e', ')', '\n',
' ', ' ', ' ', ' ', '}', '\n',
'}', '\n',
 };
```

The code was just recycling the contents of `utils/encode.c` with some minor adjustments.

### DIFF
--- a/src/util/wren2c.c
+++ b/src/util/wren2c.c
@@ -1,90 +1,55 @@
-#include <stdio.h>
-#include <string.h>
-#include <stdlib.h>
 
-char* WREN2C_hash(char *key)
-{
-    // https://en.wikipedia.org/wiki/Jenkins_hash_function
-    // http://www.burtleburtle.net/bob/hash/doobs.html
-    int len = strlen(key);
-    uint32_t hash, i;
-    for(hash = i = 0; i < len; ++i)
-    {
-        hash += key[i];
-        hash += (hash << 10);
-        hash ^= (hash >> 6);
+internal 
+void WREN2C_randomString(char *dest) {
+    // https://stackoverflow.com/a/15768317
+    char charset[] = "0123456789"
+                     "abcdefghijklmnopqrstuvwxyz"
+                     "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+    srand(time(0));
+
+    size_t length = 10;
+    while (length-- > 0) {
+        size_t index = (double) rand() / RAND_MAX * (sizeof charset - 1);
+        *dest++ = charset[index];
     }
-    hash += (hash << 3);
-    hash ^= (hash >> 11);
-    hash += (hash << 15);
-
-    char out[10];
-    sprintf(out, "%d", hash);
-
-    return out;
-}
-
-char* WREN2C_readEntireFile(char* path, size_t* lengthPtr) {
-  FILE* file = fopen(path, "r");
-  if (file == NULL) {
-    return NULL;
-  }
-  char* source = NULL;
-  if (fseek(file, 0L, SEEK_END) == 0) {
-    /* Get the size of the file. */
-    long bufsize = ftell(file);
-    /* Allocate our buffer to that size. */
-    source = malloc(sizeof(char) * (bufsize + 1));
-
-    /* Go back to the start of the file. */
-    if (fseek(file, 0L, SEEK_SET) != 0) { /* Error */ }
-
-    /* Read the entire file into memory. */
-    size_t newLen = fread(source, sizeof(char), bufsize, file);
-    if ( ferror( file ) != 0 ) {
-      fputs("Error reading file", stderr);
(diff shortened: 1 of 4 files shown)


src/util/wren2c.c
```diff
@@ -0,0 +1,81 @@
...
+
+
+internal 
+void WREN2C_encodeAndDump(ENGINE* engine, int argc, char* args[]) {
+  
+  char* fileName = args[2];
+  char* fileOut = strcat(strdup(fileName), ".inc");
+
+  char randName[] = "";
+  WREN2C_randomString(randName);
```
avivbeeri, line 27: Why are you using a random name here? It would be better to allow a consistent and user-set module/variable name so this can be incorporated into build tools.

src/util/wren2c.c
```diff
@@ -0,0 +1,81 @@
...
+
+  char* moduleName = randName;
```
avivbeeri, line 29: It would be better to only compute a alternate/default name iff the moduleName wasn't specified. I'm not certain it should be randomised, but a default is good for testing.

src/util/wren2c.c
```diff
@@ -0,0 +1,81 @@
...
+    while (length-- > 0) {
+        size_t index = (double) rand() / RAND_MAX * (sizeof charset - 1);
+        *dest++ = charset[index];
+    }
+    *dest = '\0';
+}
+
+
+internal 
+void WREN2C_encodeAndDump(ENGINE* engine, int argc, char* args[]) {
```
avivbeeri, line 21: It would be nice if this method was shared between DOME and the `utils/embed.c` tool, to avoid the duplication.

clsource: If I remove the engine dependency, then it can be used as a stand alone function.
I just put it there for maybe using the logger. But can be ditched with no problems 👍

src/util/wren2c.c
clsource: Having the example https://github.com/wren-lang/wren/blob/main/util/wren_to_c_string.py it uses the filename as the default module name. Maybe something similar can be used for a default name. But i dont know how complex can be done that in C.

src/util/wren2c.c
clsource: Additionally. If this command would be available using the DOME executable.
Should embed.c be removed? Since it just a single main function with no relations to other DOME internals.

avivbeeri: No, we need the embed.c available separately to bootstrap DOME's compilation. 

Engine core developers might not have DOME installed before they want to start working on DOME ;)

clsource:
As this has unwanted commits I will work on another PR

PR #155, opened 2021-02-01
extracted wren2cstring from embed.c

clsource:
better than #151

### DIFF
--- a/src/util/embed.c
+++ b/src/util/embed.c
@@ -1,37 +1,4 @@
-//Using standard IO only
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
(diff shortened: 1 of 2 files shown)


avivbeeri:
This is failing to build currently because modules are missing from the code. My guess is that the naming of the modules has changed from how they used to be named.

clsource:
> This is failing to build currently because modules are missing from the code. My guess is that the naming of the modules has changed from how they used to be named.

weird. maybe my base branch is wrong again.
I will redo the thing from scratch

avivbeeri:
You might just need to rebuild the modules in order to see the effects. If the modules didn't change it may not rebuild.

Do `make clean` before `make` and see.

PR #158, opened 2021-02-04
embed command

clsource:
The wren2cstring command that helps plugin development.
improved version of #151

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -117,7 +117,7 @@ global_variable size_t GIF_SCALE = 1;
 #include "modules/platform.c"
 #include "modules/random.c"
 #include "modules/plugin.c"
-#include "util/wren2cstring.c"
+#include "util/wren2cinc.c"
 
 
 // Comes last to register modules
@@ -349,6 +349,7 @@ printUsage(ENGINE* engine) {
 
   ENGINE_printLog(engine, "  dome [options]\n");
   ENGINE_printLog(engine, "  dome [options] [--] entry_path [arguments]\n");
+  ENGINE_printLog(engine, "  dome -w | --wren2cinc [--] sourceFile [moduleName] [destinationFile]\n");
   ENGINE_printLog(engine, "  dome -h | --help\n");
   ENGINE_printLog(engine, "  dome -v | --version\n");
   ENGINE_printLog(engine, "\nOptions: \n");
@@ -360,7 +361,7 @@ printUsage(ENGINE* engine) {
   ENGINE_printLog(engine, "  -h --help           Show this screen.\n");
   ENGINE_printLog(engine, "  -r --record=<gif>   Record video to <gif>.\n");
   ENGINE_printLog(engine, "  -v --version        Show version.\n");
-  ENGINE_printLog(engine, "  -w --wren2cstring   Converts a Wren source to a C String (helps plugin dev).\n");
+  ENGINE_printLog(engine, "  -w --wren2cinc      Converts a Wren source file to a C include file.\n");
 }
 
 int main(int argc, char* args[])
@@ -393,7 +394,7 @@ int main(int argc, char* args[])
     {"version", 'v', OPTPARSE_NONE},
     {"record", 'r', OPTPARSE_OPTIONAL},
     {"scale", 's', OPTPARSE_REQUIRED},
-    {"wren2cstring", 'w', OPTPARSE_REQUIRED},
+    {"wren2cinc", 'w', OPTPARSE_REQUIRED},
     {0}
   };
 
@@ -450,7 +451,7 @@ int main(int argc, char* args[])
         printVersion(&engine);
         goto cleanup;
       case 'w':
-        WREN2CSTRING_encodeAndDump(argc, args, WREN2CSTRING_INSIDE_DOME);
+        WREN2CINC_encodeAndDumpInDOME(argc, args);
         goto cleanup;
       case '?':
         fprintf(stderr, "%s: %s\n", args[0], options.errmsg);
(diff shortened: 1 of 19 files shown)


src/util/wren2cstring.c
```diff
@@ -34,27 +40,27 @@ char* WREN2CSTRING_readEntireFile(char* path, size_t* lengthPtr)
   return source;
 }
 
-int WREN2CSTRING_encodeAndDump(int argc, char* args[])
+int WREN2CSTRING_encodeAndDump(int argc, char* args[], int inDOME)
```
avivbeeri, line 43: As an alternative to this new parameter, you could just adjust the args pointer you pass in to the function to be one step along.

src/main.c
```diff
@@ -391,6 +393,7 @@ int main(int argc, char* args[])
     {"version", 'v', OPTPARSE_NONE},
     {"record", 'r', OPTPARSE_OPTIONAL},
     {"scale", 's', OPTPARSE_REQUIRED},
+    {"wren2cstring", 'w', OPTPARSE_REQUIRED},
```
avivbeeri, line 396: I'm not certain `OPTPARSE_REQUIRED` is correct because we don't expect an argument for this flag. `OPTPARSE_NONE` makes more sense.

src/main.c
```diff
@@ -359,6 +360,7 @@ printUsage(ENGINE* engine) {
   ENGINE_printLog(engine, "  -h --help           Show this screen.\n");
   ENGINE_printLog(engine, "  -r --record=<gif>   Record video to <gif>.\n");
   ENGINE_printLog(engine, "  -v --version        Show version.\n");
+  ENGINE_printLog(engine, "  -w --wren2cstring   Converts a Wren source to a C String (helps plugin dev).\n");
```
avivbeeri, line 363: An extra line should be added to show how this is used, under line 351
```
 352 |  ENGINE_printLog(engine, "  dome -w | --wren2cstring [--] sourceFile [moduleName] [destinationFile]\n");
```

I would change the description of the flag to "Converts a Wren source file to a C include file."

avivbeeri (changes requested):

src/main.c
clsource: Yes we expect at least the filename to be encoded :)

avivbeeri: That is true, but we aren't using optparse here to retrieve that argument.

clsource:
Changed the name to `wren2cinc` since its shorter and more descriptive

src/main.c
clsource: Is useful to check it before calling the function by optparse validation. Is a way to check before executing :)

avivbeeri:
I'm not sure I like the name of the parameter at the moment, but I've not decided on another option.

clsource:
> I'm not sure I like the name of the parameter at the moment, but I've not decided on another option.

we can also use `embed` as the utility?

avivbeeri:
I'd be happy to name it `--embed` and use `-e`

src/main.c
avivbeeri: What happens if no other arguments are provided? `dome -e`

clsource: ./dome: option requires an argument -- 'e'

src/main.c
```diff
@@ -387,10 +390,11 @@ int main(int argc, char* args[])
     {"console", 'c', OPTPARSE_NONE},
     #endif
     {"debug", 'd', OPTPARSE_NONE},
+    {"embed", 'e', OPTPARSE_REQUIRED},
```
avivbeeri, line 393: This is not sufficient. Because optparse _expects_ this to be followed by an argument, different forms of invocation can break.

For example:
```bash
./dome module.wren -e    # This will error out.
./dome module.wren -e module  # This seg faults because it tries to use "module" as a file name.
```

This should be `OPTPARSE_NONE`, and we should do the argument checks ourselves (including the actual presence of the input file.)

avivbeeri:
Please resolve the conflicts with `develop`: I had to make a small tweak because `strdup` was not available to the standalone `embed` tool.

src/main.c
clsource: ok

src/main.c
```diff
@@ -349,7 +349,7 @@ printUsage(ENGINE* engine) {
 
   ENGINE_printLog(engine, "  dome [options]\n");
   ENGINE_printLog(engine, "  dome [options] [--] entry_path [arguments]\n");
-  ENGINE_printLog(engine, "  dome -e | --embed [--] sourceFile [moduleName] [destinationFile]\n");
```
avivbeeri: The `[--]` should be kept because it is part of the usage format, even if it is optional.

src/util/wrenembed.c
```diff
@@ -114,14 +116,19 @@ int WRENEMBED_encodeAndDump(int argc, char* args[])
   return EXIT_SUCCESS;
 }
 
+void WRENEMBED_usage() {
+  fputs("dome -e | --embed sourceFile [moduleName] [destinationFile]\n", stderr);
```
avivbeeri, line 120: I guess this is only `dome` if it isn't the standalone embed tool.

src/main.c
clsource: Since we are using raw argc and args the `--` is considered a valid param for a filename. So using it will cause file not found.

src/util/wrenembed.c
clsource: Yes it is only used in the dome function. Was created only to avoid repetition.

avivbeeri: This should have `[--]` before the `sourceFile`, because it is a legal, but optional addition. The same applies to the DOME usage message.

PR #159, opened 2021-02-08
Fixed vsync saying "string" instead of "bool"

BenStigsen:


avivbeeri (approved):
Thank you. Little things like this do slip past occasionally, no matter how careful you are!

Issue #160, opened 2021-02-08
Font metaclass does not implement 'load(_,_)'


BenStigsen:
I'm not able to load a font to use for drawing. I'm not too experienced with Wren, but these are the things I've tried to load the font in the same directory as the game:
- `Font.load("monogram", "./monogram.ttf")`
- `Canvas.font = "./monogram.ttf"`

This gives the error `Font metaclass does not implement 'load(_,_)'`

I'm not sure if this is me or the docs, but none of these work. I can however see that the `load()` function in font.wren is `static` (so is `unload()`). I'm guessing this makes the functions private, so I'm not able to call them?

avivbeeri:
`static` means that they belong to the class, rather than to "instances" of the class. Every method in Wren is public.

The problem is that `load` takes three arguments, not two: The name, path and size.

You should do:
```
Font.load("monogram", "./monogram.tff", 8")
Canvas.font = "monogram"
```

BenStigsen:
Ah, thank you. _(sorry for the unnecessary PR)_

avivbeeri:
No worries, it's a learning process :) The issue with TTF fonts is that you need to give them some kind of size, otherwise I'd have put some kind of default there. 

(Plus, sizes aren't uniform across fonts)

PR #161, opened 2021-02-08
Make font loading/unloading public

BenStigsen:
#160

### DIFF
--- a/src/modules/font.wren
+++ b/src/modules/font.wren
@@ -12,7 +12,7 @@ class Font {
 
   static default { null }
 
-  static load(name, path, size) {
+  load(name, path, size) {
     if (!__fontFiles.containsKey(path)) {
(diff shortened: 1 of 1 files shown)


BenStigsen:
Should probably add a comment. Currently it's not possible to load a custom font because the class functions are private. I'm not sure if `unload()` should be public, but `load()` probably should.

avivbeeri:
This PR isn't valid, although I appreciate the work that goes into it. `static` means that the _class_ `Font` has the method, rather than individual instances of that class.

BenStigsen:
Oh, sorry about that. Yeah, I should probably just have waited for your issue reply since you responded very quickly. If I find something else I'll look properly into it first :)

avivbeeri:
I'm not always this speedy, but I usually respond within about 12 hours.

I am glad that you are engaging with the docs and source though :)

PR #167, opened 2021-02-11
Newline support for fonts

BenStigsen:
This adds newline support in for Canvas.print(). A default spacing of `fontHeight / 4` has been set. The user should probably be able to customize this at some point.

### DIFF
--- a/docs/modules/graphics.md
+++ b/docs/modules/graphics.md
@@ -60,6 +60,7 @@ Offset all following draw operations by (_x, y_). Calling this without arguments
 
 #### `static print(str, x: Number, y: Number, c: Color) `
 Print the text _str_ with the top-left corner at (_x, y_) in color _c_, using the currently set default font. See `Canvas.font` for more information.
+If a newline `\n` is in the string, the text will broken over multiple lines with a spacing of `fontHeight / 4`. _The spacing is currently not customizable, but will be later_
 
 #### `static print(str, x: Number, y: Number, c: Color, fontName: String) `
 Print the text _str_ with the top-left corner at (_x, y_) in color _c_, in the specified font.
--- a/docs/modules/graphics.md
+++ b/docs/modules/graphics.md
@@ -60,10 +60,13 @@ Offset all following draw operations by (_x, y_). Calling this without arguments
 
 #### `static print(str, x: Number, y: Number, c: Color) `
 Print the text _str_ with the top-left corner at (_x, y_) in color _c_, using the currently set default font. See `Canvas.font` for more information.
-If a newline `\n` is in the string, the text will broken over multiple lines with a spacing of `fontHeight / 4`. _The spacing is currently not customizable, but will be later_
+The text will be split across multiple lines if a newline (`\n`) character is encountered.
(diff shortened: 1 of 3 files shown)


src/engine.c
```diff
@@ -485,19 +485,28 @@ defaultFontLookup(utf8_int32_t codepoint) {
...
   int fontWidth = 8;
   int fontHeight = 8;
+  int spacing = (fontHeight / 4);
   int cursor = 0;
   utf8_int32_t codepoint;
   void* v = utf8codepoint(text, &codepoint);
   size_t len = utf8len(text);
   for (size_t pos = 0; pos < len; pos++) {
+    if (text[pos] == '\n') {
+      newlines++;
```
avivbeeri, line 498: We can just modify the `y` value directly, instead of counting the newlines.

avivbeeri (changes requested):
Generally looks good.

One small tweak. Could you also add this to the documentation, the correct file is: `docs/modules/graphics.md`

src/engine.c
BenStigsen: Ah yeah that's true

BenStigsen:
It should be ready for merging.

BenStigsen:
It should now be ready for merging.

src/modules/font.c
```diff
@@ -82,6 +87,9 @@ FONT_RASTER_print(WrenVM* vm) {
   unsigned char *bitmap;
   int w, h;
 
+  int fontHeight = raster->height;
+  int newlines = 0;
```
avivbeeri, line 91: We shouldn't need a newLines variable here, for the same reason as last time.

BenStigsen: In this case for rasterized fonts the height can't be added directly to `posY` since they are calculated further down, with variables initialized later. This could be done if the `bitmap` was calculated even though we run into newlines, but that would be a waste of performance since 4 variables are initialized and two functions are called which don't need to be called.

docs/modules/graphics.md
```diff
@@ -60,6 +60,7 @@ Offset all following draw operations by (_x, y_). Calling this without arguments
 
 #### `static print(str, x: Number, y: Number, c: Color) `
 Print the text _str_ with the top-left corner at (_x, y_) in color _c_, using the currently set default font. See `Canvas.font` for more information.
+If a newline `\n` is in the string, the text will broken over multiple lines with a spacing of `fontHeight / 4`. _The spacing is currently not customizable, but will be later_
```
avivbeeri, line 63: I think the docs should read something more like this:

> The text will be split across multiple lines if a newline (`\n`) character is encountered. 
> The vertical line spacing defaults to `fontHeight / 4`, where `fontHeight` is the maximum height of a font character in pixels, but this will be configurable in future versions of DOME.

src/modules/font.c
avivbeeri: I believe in these case we would actually modify the value of the `baseY` rather than the `posY` value, as that is the Y position for the baseline of the text.

BenStigsen: Ah yeah that's true, I didn't even consider that.

avivbeeri:
All looks good! I remembered as I was merging this that it is recommended that you add your name/email/whatever to the AUTHORS.md file, as a record of your contribution! Feel free to submit an additional PR for that, or directly message me what you'd like :)

PR #194, opened 2021-04-12
Emscripten support

avivbeeri:
I'd like to integrate this in sooner than later because of other refactoring projects ongoing.

DOME won't support web as a first-class target (due to serious performance issues), but it's an option for more lightweight projects.

Documentation of some kind is still required, as is generating the web-runner.


avivbeeri:
There's also no fullscreen support at the moment because for some reason the resizing viewport doesn't play nicely with mouse positioning.

PR #200, opened 2021-05-01
Added Canvas.font getter

BenStigsen:
This makes it possible to do `Canvas.font` to get the current font name as a string.

### DIFF
--- a/docs/modules/graphics.md
+++ b/docs/modules/graphics.md
@@ -18,12 +18,12 @@ It contains the following classes:
 The `Canvas` class is the core api for graphical display.
 
 ### Fields
+#### `static font: String`
+This is the name of the default font used for `Canvas.print(str, x, y, color)`. You can set this to `Font.default` to return to the DOME built-in font.
 #### `static height: Number`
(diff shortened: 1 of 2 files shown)


src/modules/graphics.wren
```diff
@@ -13,6 +13,8 @@ class Canvas {
       Fiber.abort("Default font must be a font name")
     }
   }
+  
+  static font {__defaultFont}
```
avivbeeri, line 17: Slight style nitpick, I prefer to have spaces between braces and the statement within.

This may also need an update to the documentation to make it clear this field can be both written to and read.

docs/modules/graphics.md
```diff
@@ -22,6 +22,8 @@ The `Canvas` class is the core api for graphical display.
 This is the height of the canvas/viewport, in pixels.
 #### `static width: Number`
 This is the width of the canvas/viewport, in pixels.
+#### `static font: String`
```
avivbeeri, line 25: Oops, There's actually already documentation for this field, it's just in the wrong place. Let's move it up to here (and preferably keep the fields in alphabetical order).

PR #207, opened 2021-08-29
Added Squirrel5 pseudo RNG

TheKing0x9:
This pull adds Squirrel5 pseudo RNG, accessible via the `squirrel5` module. The functionality of this module has been kept at par with the current `squirrel3` RNG.

Please check if the modifications made to `main.c` and `generateEmbedModules.sh` are sufficient to add a new module to Dome.


avivbeeri:
I would prefer this to be part of the Random module, but just as a separate distinct class. Separating them out isn't very useful.

TheKing0x9:
Apologies for the late reply, but since both of them have the exact same functions, how should I go about structuring the documentation. Should I provide the documentation for both of them seperately, or explain both of them with their differences at the start, and then state that `both of them have these common functionalities` followed by the documentation of the members?

avivbeeri:
There shouldn't be any particular differences, beyond explaining that DOME
has two options for the random algorithm, so that's the approach I would
take.

"Squirrel3 (default) and Squirrel5 both provide the same API, which is
explained below:"

TheKing0x9:
Fixed!

avivbeeri (approved):

avivbeeri:
Thank you for doing this :)

Issue #203, opened 2021-05-21
Error dialogs can be annoying during development


avivbeeri:
Some developer environments work better with consoles than others, so being harassed by error dialogs can lead to frustration and interruptions in development flow.

It would be good to have a global setting to turn them off. `Process.errorDialogs` as a field perhaps?

PR #210, opened 2021-10-08
Added Process.errorDialogue

TheKing0x9:
This pull adds Process.errorDialogue which allows errors message boxes to be disabled and instead logged only in the console. ( #203 )


avivbeeri:
Actually, I need to change the API here slightly to be `Process.errorDialog`

PR #218, opened 2021-11-23
added Mouse.setCursor to set the system cursor through SDL

trelemar:


src/engine.c
```diff
@@ -881,6 +881,24 @@ ENGINE_getKeyState(ENGINE* engine, char* keyName) {
...
+    if (strcmp(name, cursorName) == 0) cursorIndex = index;
+  }
+  return cursorIndex;
+}
+
+internal void
+ENGINE_setMouseCursor(ENGINE* engine, char* cursorName) {
+  SDL_FreeCursor(engine->mouse.cursor);
+  int cursorIndex = ENGINE_findMouseCursorIndex(engine, cursorName);
+  engine->mouse.cursor = SDL_CreateSystemCursor(cursorIndex);
```
avivbeeri, line 898: We need to `SDL_FreeCursor` the previous cursor before doing this. (To do that safely, we should also initialise this property to NULL). We also need to do the free during `ENGINE_free`.

avivbeeri: It'd be nice I think if we could cache this value, but we don't have the infrastructure for it at the moment.

docs/modules/input.md
```diff
@@ -127,6 +127,22 @@ You can use an index from 1-5 (button 0 is invalid) or a lowercase name:
...
+* `wait`
+* `crosshair`
+* `waitarrow`
+* `sizenwse`
+* `sizenesw`
+* `sizewe`
+* `sizens`
+* `sizeall`
+* `no`
+* `hand`
```
avivbeeri, line 144: Where did these names come from? I think I'd like to consider how we name them a bit more carefully, if I can make a choice :P

src/engine.c
```diff
@@ -881,6 +881,24 @@ ENGINE_getKeyState(ENGINE* engine, char* keyName) {
   return state[scancode];
 }
 
+internal uint32_t
+ENGINE_findMouseCursorIndex(ENGINE* engine, char* cursorName) {
+  int cursorIndex;
+  for (int index = 0; index < 12; index++) {
```
avivbeeri, line 887: SDL Enums usually have an entry at the end which can be used as the upper limit of entries. See here: https://github.com/libsdl-org/SDL/blob/e18be04bc6ae964beb175b5468595d640a100eec/include/SDL_mouse.h#L60

So we should use `SDL_NUM_SYSTEM_CURSORS` instead of 12. :)

src/engine.c
```diff
@@ -881,6 +881,24 @@ ENGINE_getKeyState(ENGINE* engine, char* keyName) {
...
+    char * name = ENGINE_MOUSE_CURSORS[index];
+    if (strcmp(name, cursorName) == 0) cursorIndex = index;
```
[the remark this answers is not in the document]
avivbeeri: This should probably be:
```c
internal int32_t
ENGINE_findMouseCursorIndex(ENGINE* engine, char* cursorName) {
  for (int index = 0; index < SDL_NUM_SYSTEM_CURSORS; index++) {
    char * name = ENGINE_MOUSE_CURSORS[index];
    if (STRINGS_EQUAL(cursorName, name)) {
      return index;
    }
  }
  return -1;
}
```

There's always the possibility that an invalid name is provided, so we need to be able to detect and handle that error case. (Note that I changed the returntype to be an `int32_t` so that we have an error option.)

avivbeeri (changes requested):
This is an interesting PR. I still need to look a bit more at how the Cursor API works for SDL to decide if this implementation is as good as it can be.

src/modules/input.wren
```diff
@@ -115,6 +115,7 @@ class Mouse {
   foreign static hidden=(value)
   foreign static relative=(value)
   foreign static relative
+  foreign static setCursor(name)
```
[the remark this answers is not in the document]
trelemar: I agree. After working with the method on a practical case, It it makes sense to have a getter. It would be better if it used that signature as well. I'll look into implementing that.

src/modules/input.c
```diff
@@ -254,6 +254,14 @@ MOUSE_getHidden(WrenVM* vm) {
   wrenSetSlotBool(vm, 0, !shown);
 }
 
+internal void
+MOUSE_setCursor(WrenVM* vm) {
+  ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
+  ASSERT_SLOT_TYPE(vm, 1, STRING, "cursorName");
+  char* name = (char*)wrenGetSlotString(vm, 1);
```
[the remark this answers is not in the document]
trelemar: fixed in 508b30cd61060968035a901f0147c8465a6c9129

docs/modules/input.md
trelemar: These names are just the tail end of the enum definitions from SDL in lowercase. We can name them anything :)
https://wiki.libsdl.org/SDL_SystemCursor

src/engine.c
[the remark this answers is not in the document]
trelemar: Currently if an unknown name is provided the loop does not change `cursorIndex` so it will return 0. So if you try to set the cursor to an unknown name it will set it back to the default arrow cursor. It's your call if you want the method to throw an error if that makes more sense tho.

src/engine.c
trelemar: So when I'm calling `SDL_FreeCursor(engine->mouse.cursor);` you're saying it's not actually doing anything? Or should i be calling that somewhere else?

src/engine.c
trelemar: fixed in 8d7232c

src/engine.c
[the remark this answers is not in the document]
trelemar: updated your changes in 8d7232c, still no error checking but SDL seems to take care of it either way.

src/engine.c
avivbeeri: Oh..no, I'm just blind, turns out. Sorry, I didn't see the free there first time around. We need to call it in ENGINE_free though.

src/modules/input.wren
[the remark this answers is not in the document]
trelemar: fixed in a028707

src/engine.c
trelemar: No worries! Fixed in 8613623

avivbeeri (commented):
I'm happy with the implementation now. I'm not 100% sure yet if the text-based names need to be changed or not. I would need to experiment with them and see.

We may want to consider introducing an "Enum" for them on the Wren side.

trelemar:
> I'm not 100% sure yet if the text-based names need to be changed or not. I would need to experiment with them and see.
> 
> We may want to consider introducing an "Enum" for them on the Wren side.

I can implement any change once you test it out. Just let me know 👌

docs/modules/input.md
avivbeeri: Happy with these names, although not all the cursors are available on every platform (mostly OSX is the problem here) so we should make a note of that in the documentation.

trelemar: I added some notes to the docs about what cursors are available on mac in 570a0a1. If you'd like to make any further changes feel free :smile:

avivbeeri: I'm debating whether we should specifically say what OSX does, or just be more general, to cover other platforms "These cursors may not be available on all platforms, or may display differently."

avivbeeri (approved):

PR #219, opened 2021-11-23
aarch64 support on Linux

trelemar:


avivbeeri:
I said this on discord already, but for transparency:

I think the proper approach for this is to swap to using Wren's amalgamated build. We should commit a version of it to the repo, and then we can re-generate it when Wren releases a new version. This can then be built into its own module (similar to the vendor module), and linked in.

trelemar:
Closed in favor of amalgamated wren build #220

PR #220, opened 2021-12-01
Use amalgamated build of wren for wider range of device compatibility

trelemar:
@avivbeeri  #219
>I think the proper approach for this is to swap to using Wren's amalgamated build. We should commit a version of it to the repo, and then we can re-generate it when Wren releases a new version. This can then be built into its own module (similar to the vendor module), and linked in.

This pr also disables using the glibc_compat on arm64 linux as the build currently fails without doing that. This is a temporary workaround until a better solution is found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,11 +143,6 @@ SDLFLAGS=$(shell $(SDL_CONFIG) --libs)
 endif
 endif
 
-ifneq ($(filter release,$(TAGS)),)
-#DEPS += -lwren
-else ifneq ($(filter debug,$(TAGS)),)
-#DEPS += -lwrend
-endif
 ifneq ($(and $(filter windows,$(TAGS)),$(filter static,$(TAGS))),)
 WINDOW_MODE ?= windows
 WINDOW_MODE_FLAG = -m$(WINDOW_MODE)
@@ -173,7 +168,7 @@ PROJECTS := dome.bin modules
 
 all: $(PROJECTS)
 WREN_LIB ?= $(OBJS)/libwren.o
-WREN_PARAMS ?= $(ARCH) WREN_OPT_RANDOM=0 WREN_OPT_META=1
+WREN_PARAMS ?= -DWREN_OPT_RANDOM=0 -DWREN_OPT_META=1
 
 $(TOOLS)/embed: $(TOOLS)/embed-standalone.c $(TOOLS)/embedlib.c
 	@echo "==== Building standalone embed tool  ===="
@@ -185,10 +180,10 @@ $(MODULES)/*.inc: $(TOOLS)/embed $(MODULES)/*.wren
 
 modules: $(MODULES)/*.inc
 
-$(OBJS)/libwren.o:
+$(OBJS)/libwren.o: $(INCLUDES)/wren.c
 	@mkdir -p $(OBJS)
 	@echo "==== Building wren module ===="
-	$(CC) -c $(INCLUDES)/wren.c -o $(OBJS)/libwren.o $(IFLAGS)
+	$(CC) -c $(INCLUDES)/wren.c -o $(OBJS)/libwren.o $(WREN_PARAMS) $(IFLAGS)
 
 $(OBJS)/glibc_compat.o: $(INCLUDES)/glibc_compat.c
 	@mkdir -p $(OBJS)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -188,6 +185,11 @@ $(MODULES)/*.inc: $(TOOLS)/embed $(MODULES)/*.wren
 
 modules: $(MODULES)/*.inc
 
+$(OBJS)/libwren.o:
+	@mkdir -p $(OBJS)
+	@echo "==== Building wren module ===="
+	$(CC) -c $(INCLUDES)/wren.c -o $(OBJS)/libwren.o $(IFLAGS)
```
avivbeeri, line 191: Do we need to pass the WREN_PARAMS in here, to ensure we get the correct optional modules?

Makefile
```diff
@@ -19,13 +19,18 @@ MODE ?= release
...
+UNAME_M = $(shell uname -m)
 ifeq ($(UNAME_S), Darwin)
 SYSTEM ?= macosx
 ARCH ?= 64bit
 FRAMEWORK ?= $(shell which sdl2-config 1>/dev/null && echo "" || echo "framework")
 else ifeq ($(UNAME_S), Linux)
 SYSTEM ?= linux
-ARCH ?= 64bit
+	ifeq ($(UNAME_M), aarch64)
+		ARCH ?= arm64
```
avivbeeri, line 30: Do we still need to detect the architecture?

Makefile
trelemar: Fixed in 4847c6c

Makefile
```diff
@@ -139,9 +144,9 @@ endif
 endif
 
 ifneq ($(filter release,$(TAGS)),)
-DEPS += -lwren
+#DEPS += -lwren
 else ifneq ($(filter debug,$(TAGS)),)
-DEPS += -lwrend
+#DEPS += -lwrend
```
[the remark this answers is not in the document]
trelemar: Fixed in 4847c6c

Makefile
trelemar: To disable the glibc_compat on arm64 I do until a better solution is found. I haven't looked into an alternative yet.

avivbeeri: Ah I see. No worries then :)

Makefile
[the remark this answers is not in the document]
avivbeeri: We need to add a `-DDEBUG` flag to the Wren params if we are building in debug mode, so that we can switch between the release and debug modes of the Wren VM. (This is only for engine development really, but still important to me :P

Makefile
```diff
@@ -167,16 +167,8 @@ PROJECTS := dome.bin modules
...
-WREN_PARAMS ?= $(ARCH) WREN_OPT_RANDOM=0 WREN_OPT_META=1   
-$(LIBS)/wren/lib/libwren.a:
-	@echo "==== Cloning Wren ===="
-	git submodule update --init -- $(LIBS)/wren
-$(LIBS)/wren: $(LIBS)/wren/lib/libwren.a
-$(WREN_LIB): $(LIBS)/wren
-	@echo "==== Building Wren ===="
-	./scripts/setup_wren.sh $(WREN_PARAMS)
+WREN_LIB ?= $(OBJS)/libwren.o
+WREN_PARAMS ?= -DWREN_OPT_RANDOM=0 -DWREN_OPT_META=1
```
avivbeeri, line 171: I'd probably move these to near the top of the file, and then we can append `-DDEBUG=1` to the `WREN_PARAMS` variable if we are compiling in debug mode.

avivbeeri (changes requested):
Just need the debug flag and then this looks great :)

trelemar:
Fixed the debug parameter in 7d03937.

We don't need to explicitly state DEBUG=0 if compiling the release mode right? For clarity we could but I left it out.

Makefile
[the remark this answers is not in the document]
trelemar: Fixed in 7d03937 & 179456c

Makefile
```diff
@@ -112,7 +113,6 @@ ifneq ($(filter release,$(TAGS)),)
 CFLAGS += -O3
 else ifneq ($(filter debug,$(TAGS)),)
 CFLAGS += -O0
-WREN_PARAMS += -DDEBUG=1
```
avivbeeri: Oh, I didn't realise this line was here. In which case we may be alright, regardless of this change. I should have double-checked.

trelemar: Because It wasn't there the last time you looked :laughing:  I just added it in 7d03937 but changed it in 179456c. I think it makes sense where it is now.

avivbeeri (approved):

avivbeeri:
Happy to merge this, if there's nothing else to adjust. :)

trelemar:
Looks good to me! Thank you.

PR #237, opened 2022-05-06
Add method to FileSystem to check for file existance

8-BIT-DEV:
I saw in the discord server someone was asking about this and though it was A simple enough feature to implement. So I did. The new method is called fileExists in the class FileSystem. It returns true if the file exists and false it not.

### DIFF
--- a/docs/modules/io.md
+++ b/docs/modules/io.md
@@ -35,3 +35,13 @@ This is a blocking operation, and so execution will stop while the file is saved
 
 #### `static createDirectory(path: String): Void`
 Given a valid `path`, creates the directory, if it doesn't already exist and makes parent directories as needed.
+
+#### `static doesFileExist(path: String): Boolean`
+Checks if the path to the given file exists.
+
+#### `static createDirectory(path: String): Number`
+Checks if the path to the given directory exists.
+Return values
+* 0 means the directory does not exist
+* 1 means the directory exists
+* 2 means the path leads to A file
--- a/docs/modules/io.md
+++ b/docs/modules/io.md
@@ -37,11 +37,8 @@ This is a blocking operation, and so execution will stop while the file is saved
 Given a valid `path`, creates the directory, if it doesn't already exist and makes parent directories as needed.
 
 #### `static doesFileExist(path: String): Boolean`
-Checks if the path to the given file exists.
+Checks if the path to the given file exists. Does not check the tar bundle
 
-#### `static createDirectory(path: String): Number`
-Checks if the path to the given directory exists.
-Return values
-* 0 means the directory does not exist
-* 1 means the directory exists
-* 2 means the path leads to A file
+#### `static doesDirectoryExist(path: String): Boolean`
+Checks if the path to the given directory exists. Will also return false if the path is to A file instead.
+Does not check the tar bundle.
--- a/src/engine.c
+++ b/src/engine.c
@@ -135,13 +135,6 @@ ENGINE_fileExists(ENGINE* engine, const char* path) {
    strcpy(pathBuf, path);
  }
  
- if (engine->tar != NULL) {
-   bool tarCheck = fileExistsInTar(engine->tar, pathBuf);
-   if (tarCheck) {
-     return true;
-   }
- }
- 
  if (path[0] != '/') {
    strcpy(pathBuf, BASEPATH_get());
    strcat(pathBuf, path);
@@ -157,17 +150,6 @@ ENGINE_directoryExists(ENGINE* engine, const char* path) {
    strcpy(pathBuf, path + 2);
  } else {
    strcpy(pathBuf, path);
- }
- 
- if (engine->tar != NULL) {
-   int tarCheck = directoryExistsInTar(engine->tar, pathBuf);
-   if (tarCheck) {
-     return 1;
-   }
-   bool fileTarCheck = fileExistsInTar(engine->tar, pathBuf);
-   if (fileTarCheck) {
-     return 2;
-   }
  }
  
  if (path[0] != '/') {
--- a/src/engine.c
+++ b/src/engine.c
@@ -125,6 +125,66 @@ ENGINE_readFile(ENGINE* engine, const char* path, size_t* lengthPtr, char** reas
   return readEntireFile(pathBuf, lengthPtr, reason);
 }
 
+internal bool
+ENGINE_fileExists(ENGINE* engine, const char* path) {
+ char pathBuf[PATH_MAX];
+
+ if (strncmp(path, "./", 2) == 0) {
+   strcpy(pathBuf, path + 2);
+ } else {
+   strcpy(pathBuf, path);
+ }
+ 
+ if (engine->tar != NULL) {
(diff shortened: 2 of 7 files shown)


avivbeeri (changes requested):
Thank you for the contribution. It isn't quite where I'd like it for merging, but it definitely gets close.

Documentation for the new method will also be required before I can merge this.

src/modules/io.wren
```diff
@@ -14,6 +14,7 @@ class FileSystem {
   foreign static prefPath(org, app)
   foreign static basePath()
   foreign static createDirectory(path)
+  foreign static fileExists(path)
```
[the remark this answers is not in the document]
8-BIT-DEV: Change the method name to doesFileExist and added doesDirectoryExist

src/modules/io.c
```diff
@@ -310,3 +310,22 @@ FILESYSTEM_createDirectory(WrenVM *vm) {
...
 }
+
+internal void
+FILESYSTEM_fileExists(WrenVM* vm) {
+	ASSERT_SLOT_TYPE(vm, 1, STRING, "file path");
+	const char* path = wrenGetSlotString(vm, 1);
+	ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
+	
+	size_t length;
+	char* data = ENGINE_readFile(engine, path, &length, NULL);
```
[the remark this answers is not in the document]
8-BIT-DEV: Implemented new functions ENGINE_fileExists and ENGINE_directoryExists

src/modules/io.c
```diff
@@ -310,3 +310,22 @@ FILESYSTEM_createDirectory(WrenVM *vm) {
     VM_ABORT(vm, "Directory could not be created");
   }
 }
+
+internal void
+FILESYSTEM_fileExists(WrenVM* vm) {
```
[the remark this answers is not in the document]
8-BIT-DEV: fixed the styling in all functions

src/modules/io.c
```diff
@@ -310,3 +310,32 @@ FILESYSTEM_createDirectory(WrenVM *vm) {
...
+ const char* path = wrenGetSlotString(vm, 1);
+ ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
+ 
+ bool fileCheck = ENGINE_fileExists(engine, path);
+ 
+ if (!fileCheck) {
+   wrenSetSlotBool(vm, 0, false);
+   return;
+ }
+ wrenSetSlotBool(vm, 0, true);
```
avivbeeri, line 327: We can replace all of this with:
```
wrenSetSlotBool(vm, 0, !fileCheck);
```

src/engine.c
```diff
@@ -125,6 +125,66 @@ ENGINE_readFile(ENGINE* engine, const char* path, size_t* lengthPtr, char** reas
...
+ENGINE_directoryExists(ENGINE* engine, const char* path) {
+ char pathBuf[PATH_MAX];
+
+ if (strncmp(path, "./", 2) == 0) {
+   strcpy(pathBuf, path + 2);
+ } else {
+   strcpy(pathBuf, path);
+ }
+ 
+ if (engine->tar != NULL) {
```
avivbeeri, line 162: I changed my mind about how this should work.
Because the tar bundle is considered read-only, I don't think we need to check in the tar for existence before the disk. 'You' as the developer should know if a file or directory exists in there. If we want to check that in the future anyway (to handle modding or something) we can add that as an option later.

avivbeeri (changes requested):

docs/modules/io.md
```diff
@@ -35,3 +35,13 @@ This is a blocking operation, and so execution will stop while the file is saved
 
 #### `static createDirectory(path: String): Void`
 Given a valid `path`, creates the directory, if it doesn't already exist and makes parent directories as needed.
+
+#### `static doesFileExist(path: String): Boolean`
+Checks if the path to the given file exists.
+
+#### `static createDirectory(path: String): Number`
```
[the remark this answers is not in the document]
8-BIT-DEV: Fixed the name and changed the method to return A boolean instead.

src/modules/io.c
8-BIT-DEV: Fixed this too changed to the smaller code

src/engine.c
8-BIT-DEV: Removed the tar bundle checks now directly checks the file system.

8-BIT-DEV:
Oops accidentally committed the changes to the wrong branch again. Fixing that!

src/modules/io.c
```diff
@@ -310,3 +310,32 @@ FILESYSTEM_createDirectory(WrenVM *vm) {
...
+ ASSERT_SLOT_TYPE(vm, 1, STRING, "dir path");
+ const char* path = wrenGetSlotString(vm, 1);
+ ENGINE* engine = (ENGINE*)wrenGetUserData(vm);
+ 
+ int dirCheck = ENGINE_directoryExists(engine, path);
+ if (dirCheck != 1) {
+   wrenSetSlotBool(vm, 0, false);
+   return;
+ }
+ wrenSetSlotBool(vm, 0, true);
```
avivbeeri, line 339: We can replace this with:
```
wrenSetSlotBool(vm, 0, dirCheck);
```

src/engine.c
```diff
@@ -125,6 +125,48 @@ ENGINE_readFile(ENGINE* engine, const char* path, size_t* lengthPtr, char** reas
...
+ 
+ if (isDirectory(pathBuf)) {
+   return 1;
+ }
+ 
+ if (doesFileExist(pathBuf)) {
+   return 2;
+ }
+ return 0;
+}
```
avivbeeri, line 168: These two functions are very similar. If possible, we should merge the common code into a single function. `ENGINE_directoryExists` and `ENGINE_fileExists` could then call that common function, and interpret the results as appropriate :) 

Less duplication of code means less code that has to be maintained, and shipped to users.

src/io.c
```diff
@@ -179,6 +179,83 @@ readFileFromTar(mtar_t* tar, char* path, size_t* lengthPtr, char** data) {
   return err;
 }
 
+internal bool
+fileExistsInTar(mtar_t* tar, char* path) {
```
avivbeeri, line 183: This and `directoryExistsInTar` might be useful in the future, but they aren't used currently, so we should remove them. We can always add them back later.

avivbeeri (changes requested):
There's a few final changes to make, but I'm happier with how this looks, at least from the API design side.

Please also add yourself to the AUTHORS file, so you can be appropriately credited :)

src/engine.c
8-BIT-DEV: Replaced all that code with 2 function calls to functiosn in src/io.c

src/io.c
8-BIT-DEV: Removed the 2 functions

avivbeeri (approved):

avivbeeri:
Thank you so much for this contribution (and putting up with my review process!)

PR #238, opened 2022-07-15
Fixes a bug in the circle example

ConfusedSky:
I know this is the smallest of changes, but I was looking through the example and got an unexpected crash due to the circle's size being negative.

So I thought I would go ahead and create a PR to fix that.


avivbeeri:
Thank you for this, I'll investigate it soon :)

PR #240, opened 2023-01-01
Added rust example

clsource:
some simple plugin made with Rust


clsource:
https://github.com/NinjasCL/dome-rust-plugin-example

PR #245, opened 2023-02-18
MP3 and FLAC Support

avivbeeri:


src/modules/audio.c
```diff
@@ -70,6 +99,8 @@ AUDIO_allocate(WrenVM* vm) {
     SDL_FreeWAV((uint8_t*)tempBuffer);
   } else if (data->audioType == AUDIO_TYPE_OGG) {
     free(tempBuffer);
+  } else if (data->audioType == AUDIO_TYPE_MP3) {
+    free(tempBuffer);
   }
```
Curanos, line 104: Shouldn't there a case for FLAC, too?

avivbeeri: You're right. I was just adding it!

Curanos: Ah yes, just a couple of minutes ago :-)
Sorry about that.

avivbeeri: Not at all, I appreciate the oversight. I wrote/merged a lot of code today so mistakes are likely.

Issue #236, opened 2022-04-21
Scale dome.html viewport


whmsft:
This sounds reasonable to be able to scale/stretch the viewport/canvas of the experimental dome.html

avivbeeri:
The reason this doesn't work right now is that SDL behaves very strangely when the canvas is resized. If I remember correctly, it would desynchronize the mouse position from the view, leading to weird input behaviour, so for the moment, dynamic scaling isn't supported. 

You can set the "window size" during your Game.init(), but that size is then fixed for the duration of the game session.

Perhaps this will improve in the future?

whmsft:
And can I get the screen max height and width?

avivbeeri:
Not right now, but I'll make a note to add that in the future.

avivbeeri:
The new web engine features a scalable web canvas, which will allow the browser to slot it into whatever space you need. 
Therefore, I'm closing this issue.

PR #262, opened 2023-03-12
Hotfixing

### LINKED ISSUE #259: Log level can't be set to "debug"
Turns out that you can't set the logging level to debug, it'll only go to INFO.

I know where the issue here is, but if someone wants to submit a fix before I do, go for it.

avivbeeri:
Fixes #259 as well as sorts out a problem with High DPI in the display.
A test solution for #236 is in here as well, as demonstrated on https://avivbeeri.itch.io/acolytes-pledge

### DIFF
--- a/main.wren
+++ b/main.wren
@@ -1,10 +1,8 @@
-import "dome" for Window
 import "graphics" for Canvas, Color
+
 class Main {
   construct new() {}
-  init() {
-    Window.resize(1920, 1280)
-  }
+  init() {}
   update() {}
   draw(dt) {
     Canvas.print("DOME Installed Successfully.", 10, 10, Color.white)
--- a/main.wren
+++ b/main.wren
@@ -1,7 +1,10 @@
+import "dome" for Window
 import "graphics" for Canvas, Color
 class Main {
   construct new() {}
-  init() {}
+  init() {
+    Window.resize(1920, 1280)
+  }
   update() {}
   draw(dt) {
     Canvas.print("DOME Installed Successfully.", 10, 10, Color.white)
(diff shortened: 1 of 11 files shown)


src/engine.c
```diff
@@ -235,8 +235,14 @@ ENGINE_init(ENGINE* engine) {
 internal int
 ENGINE_start(ENGINE* engine) {
   int result = EXIT_SUCCESS;
-#if defined _WIN32
+
+  uint32_t windowFlags = SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE;
+#if defined (_WIN32) || defined (__MINGW32__)
   SDL_setenv("SDL_AUDIODRIVER", "directsound", true);
+  SDL_SetHint(SDL_HINT_MOUSE_RELATIVE_MODE_WARP, "1");
+  SDL_SetHint(SDL_HINT_VIDEO_HIGHDPI_DISABLED, "1");
```
avivbeeri, line 243: Should remove these, they don't help

src/modules/dome.c
```diff
@@ -70,6 +70,26 @@ WINDOW_resize(WrenVM* vm) {
...
+
+  int w, h;
+  w = dm.w;
+  h = dm.h;
+
+  SDL_SetWindowSize(engine->window, min(width, w-1), min(height, h - 1));
+  int32_t newWidth, newHeight;
+  SDL_GetRendererOutputSize(engine->renderer, &newWidth, &newHeight);
+  SDL_SetWindowSize(engine->window, newWidth, newHeight);
+#else
```
avivbeeri, line 91: This needs more testing, but it might be useful to do on all platforms so that the engine can always be viewed.

main.wren
```diff
@@ -1,7 +1,10 @@
+import "dome" for Window
 import "graphics" for Canvas, Color
 class Main {
   construct new() {}
-  init() {}
+  init() {
+    Window.resize(1920, 1280)
```
avivbeeri, line 6: remove this

PR #264, opened 2023-07-29
Fix typos

kianmeng:


avivbeeri:
Thank you for doing some work on the repository. You've made changes to the project code, but also some of the header include dependencies, which weren't written by me. Even if they are typos, I'd prefer to keep the files in their original condition.

Please would you remove those from the PR?

kianmeng:
@avivbeeri Noted, I'll close the PR, since all of the typos are from the header files.

```sh
$ codespell -S ./include -L atleast,nd,te,bu
```

PR #267, opened 2023-09-24
Fix windows terminal support

Kasparsu:
Fixing windows terminal support as well as powershell and cmd support since they weren't outputing anything either.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -269,8 +269,7 @@ char* resolveEntryPath(ENGINE* engine, char* entryArgument, bool autoResolve) {
 int main(int argc, char* argv[])
 {
 #ifdef __MINGW32__
-  // WT_SESSION = Windows terminal, SESSIONNAME=Console == Powershell and CMD and TERM_PROGRAM=Tabby == Tabby
-  if(getenv("WT_SESSION") || (strcmp(getenv("SESSIONNAME"),"Console") == 0 && !getenv("SHELL")) || getenv("TERM_PROGRAM") == "Tabby") {
+  if(!GetStdHandle(STD_OUTPUT_HANDLE)) {
     if (AttachConsole(ATTACH_PARENT_PROCESS)) { 
       freopen("CONOUT$","wb",stdout);
       freopen("CONOUT$","wb",stderr);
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -268,6 +268,15 @@ char* resolveEntryPath(ENGINE* engine, char* entryArgument, bool autoResolve) {
 
 int main(int argc, char* argv[])
 {
+#ifdef __MINGW32__
+  // WT_SESSION = Windows terminal, SESSIONNAME=Console == Powershell and CMD and TERM_PROGRAM=Tabby == Tabby
```
avivbeeri, line 272: It worries me that we have to customise this for each particular terminal.

There's a chance someone tries to use some other unusual terminal and it breaks. Any chance we could do better?

Kasparsu: It doesn't seems so at the moment unless there are some way of determining wether console is allready attached or not. In anycase there is only few options for terminals on windows unlike linux or mac. Im not saying that there can't be new issues about it down the line but they either can be fixed by adding it to the list or not (if there is no env var to make it unique).As far as I know lot of users use powershell and lot use win terminal now that it comes as a default.

Kasparsu: we could add a flag that is checked at the start that also applies this fix. `./dome.exe -t` for example and you could use it with other flags then `./dome.exe -t --help` for example even if there is an issue with unknown terminal then there would be a functional workaround altho funny problem with it is that you get no output what so ever so if people don't read about this in docs they won't know about the flag like I didn't know about the -c flag until I looked into `dome-out.log` and found that it outputs `--help` in there as well.

Kasparsu: @avivbeeri is this more acceptable [887ebfc](https://github.com/domeengine/dome/pull/267/commits/887ebfc4b7de38b24586b7b12dc81bb5f0b635a7)

avivbeeri: I'll have to do a build and verify it for myself before I approve it, but that looks quite good!

Kasparsu: it do be bit wonky on cmd and ps but hey it outputs stuff so thats improvement. If run trough windows terminal its good on all 3 (ps, cmd, git bash) and on tabby it works fine as well and pure git bash works just as it did before,

avivbeeri:
Sorry to have left this hanging for so long. I intend to look at it this week, and maybe push it out as part of a patch release before the weekend.

avivbeeri:
Turns out, I'm not in a position to build this for windows right now, but it's innocuous enough that I'm happy to merge it. You should add yourself to the AUTHORS file before I do :)

PR #269, opened 2024-02-21
simplify main.wren

thacuber2a03:
Wren doesn't seem to care whether a method comes from an instance of a class or the class itself, so I made the main file more concise. it avoids the need to instantiate a class and to put an empty constructor method in it, and having the game on the class itself is more akin to how Wren works

### DIFF
--- a/main.wren
+++ b/main.wren
@@ -1,9 +1,9 @@
 import "graphics" for Canvas, Color
 
 class Game {
-  init() {}
-  update() {}
-  draw(dt) {
+  static init() {}
+  static update() {}
+  static draw(dt) {
     Canvas.print("DOME Installed Successfully.", 10, 10, Color.white)
   }
 }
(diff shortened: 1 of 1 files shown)


avivbeeri:
You're correct that Wren doesn't mind if it's static or an instance object, it only cares about methods being available.

Originally DOME shipped with a static class exactly like the one in this PR. I moved away from it across all the examples and tutorial content over the last couple years because I found that newcomers to Wren find static class syntax a little strange (mostly around fields), and it wasn't exactly clear to people what was required to use them, or how to differentiate from them.

(As a side benefit, being able to supply an instance of a class as the Game object allows for sub-frameworks that provide their own Game objects to manage the game loop, which is pretty neat for reusability. I do this for my own one, Parcel)

Thanks for the PR, all the same!

PR #277, opened 2025-04-02
Add display switching api for dome

Kasparsu:
- Add `Window.display` and `Process.displayCount`
- Add Docs
- Add Example
- Made error in previous PR where added flag for screen switching by making it boolean effectivly making it so that you can move only to screens 0 and 1. Now can open window on other screens as well.

### DIFF
--- a/docs/modules/dome.md
+++ b/docs/modules/dome.md
@@ -80,6 +80,9 @@ Returns the generic name of the operating system or platform, where possible.
 #### `time: Number`
 Returns the integer number of seconds since the unix epoch (1st January 1970).
 
+#### `displayCount: Number`
+Returns the integer number of displays on the system.
+
 ## Process
 
 ### Static Fields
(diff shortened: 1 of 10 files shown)


docs/modules/dome.md
```diff
@@ -80,6 +80,9 @@ Returns the generic name of the operating system or platform, where possible.
 #### `time: Number`
 Returns the integer number of seconds since the unix epoch (1st January 1970).
 
+#### `screenCount: Number`
```
avivbeeri: This should be displayCount, and "number of displays"

Kasparsu: It should. Whoops

Kasparsu: And now its like it never happened ;P

Kasparsu:
Anything still wrong? @avivbeeri

